PR #17, opened 2025-06-28
Fixed admin page according to comments

VladimirSushinsky:
## Github ticket

* [Github ticket](https://github.com/ita-social-projects/VictoryCenter-Back/issues/204)

## Description

Compare and eliminate inconsistencies between the mockup (Figma) and the implemented frontend functionality of the admin team page according:
https://discord.com/channels/1344044167570915339/1372089997036879955/1387397757802647642

## How it looks

image

image

## Summary of change

 - side panel black line width is 8px;
 - side panel is aligned with tabs;
 - side panel buttons have 60px width;
 - team member cards have 8px identation;
 - updated category dropdown;
 - text in photo upload button is centered.

## How to recreate changes

1. Go to admin page(http://localhost:3000/admin-page/team)
2. Open Add-member modal

## CHECK LIST

### DIFF
--- a/src/components/common/select/Select.test.tsx
+++ b/src/components/common/select/Select.test.tsx
@@ -146,6 +146,8 @@ describe('Select Component', () => {
         expect(span).toHaveStyle('color: #61615C');
 
         fireEvent.click(selectContainer);
+        expect(span).toHaveStyle('color: #061125');
+        
         const option = screen.getByText('Option 1');
         fireEvent.click(option);
 
--- a/src/components/common/select/Select.tsx
+++ b/src/components/common/select/Select.tsx
@@ -39,7 +39,7 @@ export const Select = <TValue, >({
                      }
                  }}>
         <span
-            style={selectedValue !== null && selectedValue !== undefined ? {color: "#061125"} : {color: "#61615C"}}> {selectedValue !== null && selectedValue !== undefined ? selectedValue.toString() : 'Статус'}</span>
+            style={isOpen || (selectedValue !== null && selectedValue !== undefined) ? {color: "#061125"} : {color: "#61615C"}}> {selectedValue !== null && selectedValue !== undefined ? selectedValue.toString() : 'Статус'}</span>
         <img src={isOpen ? ArrowUp : ArrowDown} alt="arrow-down"/>
         <div className={`select-options ${isOpen ? 'select-options-visible' : ''}`}>
             {options.map((opt, index) => {
--- a/src/pages/admin/team/components/team-page-toolbar/TeamPageToolbar.tsx
+++ b/src/pages/admin/team/components/team-page-toolbar/TeamPageToolbar.tsx
@@ -82,6 +82,7 @@ const ConfirmPublishModal = ({
     <Modal isOpen={isOpen} onClose={onCancel} data-testid="publish-confirm-modal">
         <Modal.Title>{TEAM_PUBLISH_NEW_MEMBER}</Modal.Title>
         <Modal.Content>
+            <></>
         </Modal.Content>
         <Modal.Actions>
             <Button onClick={onCancel} buttonStyle="secondary">
@@ -106,7 +107,7 @@ const ConfirmCloseModal = ({
     <Modal isOpen={isOpen} onClose={onCancel} data-testid="confirm-close-modal">
         <Modal.Title>{TEAM_CHANGES_LOST}</Modal.Title>
         <Modal.Content>
-
+            <></>
         </Modal.Content>
         <Modal.Actions>
             <Button onClick={onCancel} buttonStyle="secondary">
--- a/src/components/admin/admin-navigation/admin-navigation.scss
+++ b/src/components/admin/admin-navigation/admin-navigation.scss
@@ -1,6 +1,6 @@
 .admin {
   &-logo {
-    height: 110px;
+    height: 150px;
     display: flex;
     justify-content: center;
     align-items: center;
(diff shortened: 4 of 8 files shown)


src/components/common/select/Select.test.tsx
```diff
@@ -146,6 +146,8 @@ describe('Select Component', () => {
         expect(span).toHaveStyle('color: #61615C');
 
         fireEvent.click(selectContainer);
+        expect(span).toHaveStyle('color: #061125');
```
Rominos7, line 149: Suggestion:
For the future you can skip that kind of checks in tests

src/components/common/select/Select.tsx
```diff
@@ -39,7 +39,7 @@ export const Select = <TValue, >({
                      }
                  }}>
         <span
-            style={selectedValue !== null && selectedValue !== undefined ? {color: "#061125"} : {color: "#61615C"}}> {selectedValue !== null && selectedValue !== undefined ? selectedValue.toString() : 'Статус'}</span>
+            style={isOpen || (selectedValue !== null && selectedValue !== undefined) ? {color: "#061125"} : {color: "#61615C"}}> {selectedValue !== null && selectedValue !== undefined ? selectedValue.toString() : 'Статус'}</span>
```
Rominos7, line 42: Two things that I can suggest:
1) Can you refactor this part to not change styles in style tag? Reason - styles in style tag would be very hard to override with class.
What can you do is use this lib [classNames](https://www.npmjs.com/package/classnames)
And refactor this part 

```
<span
  className={classNames({ 
  'color1-class': condition 1,
  'color2-class': condition 2 })}
>
{Content}
</span>
```

2) Can you exstract 'Cтатус' string into the const and import from const file? Like `const STASTUS = 'Статус'`

src/pages/admin/team/components/team-page-toolbar/TeamPageToolbar.tsx
```diff
@@ -82,6 +82,7 @@ const ConfirmPublishModal = ({
     <Modal isOpen={isOpen} onClose={onCancel} data-testid="publish-confirm-modal">
         <Modal.Title>{TEAM_PUBLISH_NEW_MEMBER}</Modal.Title>
         <Modal.Content>
+            <></>
```
Rominos7, line 85: Just a question:
Why you need to use `<></>` here?

src/pages/admin/team/components/team-page-toolbar/TeamPageToolbar.tsx
```diff
@@ -106,7 +107,7 @@ const ConfirmCloseModal = ({
     <Modal isOpen={isOpen} onClose={onCancel} data-testid="confirm-close-modal">
         <Modal.Title>{TEAM_CHANGES_LOST}</Modal.Title>
         <Modal.Content>
-
+            <></>
```
Rominos7, line 110: Same question here

src/pages/admin/team/components/team-page-toolbar/TeamPageToolbar.tsx
VladimirSushinsky: My IDE was giving me an error here, which was annoying. Adding <></> fixed it.

src/components/common/select/Select.test.tsx
Rominos7: Its more like edge case here

You do not exlicitly set class with dedicated name in element, you more like checking color value for element

Still there is no issue if you add that check

Rominos7: Up to you:
Would be good if you set this color with a class and then check if that class (with dedicated name) is set for element

src/pages/admin/team/components/team-page-toolbar/TeamPageToolbar.tsx
Rominos7: Hmmm...
I guess it is [expects child](https://github.com/ita-social-projects/VictoryCenter-Client/blob/release/1.0.0/src/components/common/modal/Modal.tsx#L87) in `<Modal.Content>` so by adding Fragment or `<></>` basically provides "empty" child

Rominos7: For the future (not here):
We may want to consider to refactor `Modal` component so it would not be mandatory to use `Modal.Content` with empty child

src/components/common/select/Select.test.tsx
VladimirSushinsky: Understood, I will fix this later when I will be integrating classNames library.

src/components/common/select/Select.tsx
VladimirSushinsky: Fixed as suggested.

OlyaMraka (approved):

src/components/common/select/Select.tsx
```diff
@@ -39,7 +41,11 @@ export const Select = <TValue, >({
                      }
                  }}>
         <span
-            style={selectedValue !== null && selectedValue !== undefined ? {color: "#061125"} : {color: "#61615C"}}> {selectedValue !== null && selectedValue !== undefined ? selectedValue.toString() : 'Статус'}</span>
+            className={classNames('empty', {
+                'not-empty': selectedValue !== null && selectedValue !== undefined})}
```
Rominos7, line 45: One other suggestion
If you would track span state using classes `empty/not-empty` then it will be confuthin seens those two classes together like `<span class="empty not-empty">`

It would not be clear if span `is empty or not?`

you can do next
```
const isSelectedValuePresent = !!selectedValue;

<span
  className{classNames({
     'empty': !isSelectedValuePresent,
      'not-empty': isSelectedValuePresent
  })}
>
```

in that way you would have only one class present for span and it would be good for tracking span state

Issue #144, opened 2025-12-04
Add path aliases to React project


Oleh-Bashtovyi:
This task aims to replace long and fragile relative imports with short, readable alias paths. It improves project navigation, reduces the risk of broken imports during refactoring, and keeps imports consistent across the codebase.

**Example**

_Before:_

`import { Button } from '../../../shared/ui/Button';`

_After:_

`import { Button } from '@shared/ui/Button';`

PR #150, opened 2025-12-07
Fix/issue 144: Add path aliases to React project

Oleh-Bashtovyi:
## Github ticker

* [144](https://github.com/ita-social-projects/VictoryCenter-Client/issues/144)

## Description

This task aims to replace long and fragile relative imports with short, readable alias paths. It improves project navigation, reduces the risk of broken imports during refactoring, and keeps imports consistent across the codebase.

## How it looks

_Before:_

`import { Button } from '../../../shared/ui/Button';`

_After:_

`import { Button } from '@shared/ui/Button';`

## Summary of change

Add path aliases

## CHECK LIST


denys-pavskyi (approved):

MotrukOleg (approved):

craco.config.js
```diff
@@ -1,6 +1,24 @@
 const customJestConfig = require('./jest.config.js');
+const path = require('path');
 
 module.exports = {
+    webpack: {
```
snowflaki, line 5: I strongly suggest avoiding granular aliases for every top-level directory (like `@components`, `@hooks`, `@services`), as this leads to config pollution and unnecessary maintenance overhead. This approach violates the `DRY` principle: every time we add or rename a folder in src, we are forced to update `craco.config.js`, `tsconfig.json`, and the `Jest` configuration strictly to keep them in sync. It also increases cognitive load and ambiguity (for example, the overlap between `@api` and `@services`, or the need for custom names like `@app-types` to avoid conflicts).

Instead, I recommend adhering to the industry standard `Root Alias pattern`. We should configure a single alias (e.g., `@` pointing to `src`). This allows for clear, consistent imports like 
```tsx 
import { Button } from '@/components/Button' 
``` 
without requiring constant updates to the build configuration as the project scales.

snowflaki (changes requested):

MykhailoKlapchuk (approved):

craco.config.js
Rominos7: @Oleh-Bashtovyi try this (if you would follow ) `Root Alias pattern`

tsconfig.json
```
{
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "@/*": ["src/*"],  // Single root alias
    }
  }
}
```

any file in system (example)
before
`import { TEAM_CATEGORY_VALIDATION } from '../../../const/admin/team';`

after
`import { TEAM_CATEGORY_VALIDATION } from '@/const/admin/team';`

Rominos7: @snowflaki I assume this was a idea?

snowflaki: > @snowflaki I assume this was a idea?

Yeap, exactly

snowflaki (approved):

PR #404, opened 2026-05-19
Fix/issue-2177 [Single Program] "Заголовок" field does not expand according to entered text (Frame 633435)

fhyhtyh:
## Github ticker

* [Github ticket](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2177 )

## Description
This PR was created in order to fix the issue with the "Заголовок" field, which didn't expand according to entered text.

## How it looks

**Before**
image
**After**
image

## Summary of change

Bug Fix: Textarea auto-grow functionality failing on input

* Updated the `useEffect` hook in `TextAreaWithCharacterLimit.tsx` to properly calculate the required height for auto-expanding textareas.
* Added a fallback calculation for `lineHeight` to prevent `NaN` errors when the browser computes `line-height` as "normal".
* Updated the height calculation logic to explicitly account for `paddingTop`, `paddingBottom`, `borderTop`, and `borderBottom` to prevent the scrollbar from triggering prematurely.
* Adjusted the DOM reset strategy to use `height: 'auto'`, ensuring correct `scrollHeight` measurements within Flexbox wrapper components.

## How to recreate changes

1. Log in as admin
2. Go to admin panel
3. Go to programs tab
4. Open create new program modal
5. Click on the add new section button
6. Select the template illustrated below
image
7. Focus on title field ("Заголовок")
8. Enter "Contrary to popular belief, Lorem Ipsumis not simply randomd"
9. Observe the result (the field should expand to fit the text without displaying a vertical scrollbar)

## CHECK LIST

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved textarea auto-grow so height calculations reliably account for line-height fallbacks, font-size, padding and borders; max-row limits and overflow behavior are now more consistent.

* **Tests**
  * Added test coverage for auto-grow logic: line-height and font-size fallbacks, padding/border inclusion, overflow capping, uncapped max-rows, and recalculation when value or maxRows change.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ita-social-projects/VictoryCenter-Client/pull/404?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


stkossman (commented):
Great fix overall - verified locally that the auto-grow behavior works correctly with the test data from the issue.

One thing to address before merging: the new logic isn't covered by tests, and the current coverage drops below the project thresholds (despite the fact that global coverage meets its threshold):

Image

The CodeRabbit note about borders in `maxHeight` - verified locally, `borderTop` and `borderBottom` are both `0px` in this component, so it has no practical impact right now.

Also there's no need to commit new `package-lock.json` - you can remove it from this PR.

snowflaki (approved):

stkossman (approved):

PR #472, opened 2026-07-02
fix/issue-2597 [Programs] The 5th program in the dropdown list is displayed partially

yaroslavkucher:
## JIRA or Github ticker

* [2597](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2597)

## Description

Fixed a UI bug where the 5th program in the search dropdown list was displayed partially.

## Summary of change

Updated the dropdown container styles/logic to ensure exactly 5 search results are fully visible initially without being cut off.

## How to recreate changes

1. Log in as Admin.
2. Open the Programs ('Програми') tab.
3. Enter 3+ characters in the search bar.
4. Verify that exactly 5 items are fully displayed in the dropdown list.

## CHECK LIST

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the suggestions dropdown height calculation for the admin search bar, making its sizing more consistent and better matched to visible results.
  * Updated container size measurements so dropdown sizing behaves more reliably across different layout conditions.
  * Refined test coverage to better reflect real element sizing behavior.


src/hooks/common/use-container-size-from-children/useContainerSizeFromChildren.ts
mehalyna, line 1: The hook now uses getBoundingClientRect() instead of offsetHeight/offsetWidth for more precise measurements
Applies Math.ceil() to the final size

getBoundingClientRect() returns fractional pixel values, while offset* properties return integers. This provides more accuracy but could behave differently across browsers with subpixel rendering. The Math.ceil() helps ensure enough space, which is good.

This is actually an improvement for the fix.

yaroslavkucher: Could you tell me if anything needs to be changed? Or everything is fine?

mehalyna: Your call :)

mehalyna (approved):

PR #479, opened 2026-07-06
Fix/issue 2907 [Reports] "Кількість змінених життів" field validation incorrectly blocks publishing

fhyhtyh:
## Github ticket

* [Github ticket](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2907)

## Description

This PR resolves a validation issue in the Admin Panel's Media Settings tab where the "Кількість змінених життів" input field was incorrectly preventing publication. Valid numbers (e.g., 205) were triggering a "Не більше 10 символів" error. The underlying validation schema has been updated to correctly enforce a character length limit (max 10 characters) rather than a strict mathematical value limit (max value of 10).

## How it looks

https://github.com/user-attachments/assets/a2d6021c-5dd0-40d0-927a-62b2fa02a110

## Summary of change

*   **Schema Validation Fix**: Replaced the `.max()` value constraint with a custom `.test('max-length')` constraint in `reports-media-settings-schema.ts` to evaluate the character length of the input instead of its numerical size.
*   **Test Suite Updates**: Updated the unit tests in `reports-media-settings-schema.test.ts` to properly assert against string length limits (e.g., testing an 11-digit number) rather than mathematical upper bounds.

## How to recreate changes
1. Log into the Admin panel and navigate to the **"Налаштування медіа"** tab under **"Звітність"**.
2. Locate the **"Кількість змінених життів"** input field.
3. Type a number greater than 10 but fewer than 11 digits (e.g., `205`).
4. Observe that the *"Не більше 10 символів"* error is no longer displayed beneath the input.
5. Click the **"Опублікувати"** (Publish) button.
6. Verify that the confirmation popup *"Опублікувати зміни?"* successfully appears with the **"ТАК"** and **"НІ"** buttons.

## CHECK LIST

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for the “changed lives” field in report media settings, with corrected handling of numeric upper-bound (“max”) scenarios while keeping the same error messaging.
- **Tests**
  - Updated and strengthened unit tests to cover “at max” and “exceeds max” boundary cases more precisely using adjusted numeric inputs.

### DIFF
--- a/src/validation/admin/reports-schema/reports-media-settings/reports-media-settings-schema.test.ts
+++ b/src/validation/admin/reports-schema/reports-media-settings/reports-media-settings-schema.test.ts
@@ -152,8 +152,10 @@ describe('reports-media-settings-schema', () => {
                 expect(result).toBe(REPORTS_TEXT.MESSAGE.INVALID_VALUE);
             });
 
-            it('should return max error when value exceeds max', () => {
-                const overMax = REPORTS_MEDIA_SETTINGS_CHANGED_LIVES_VALIDATION.changedLives.max + 1;
+            it('should return max error when value length exceeds max', () => {
+                const overMax = Number(
+                    '1'.repeat(REPORTS_MEDIA_SETTINGS_CHANGED_LIVES_VALIDATION.changedLives.max + 1),
+                );
                 const result = REPORTS_CHANGED_LIVES_VALIDATION_FUNCTIONS.validateChangedLives(overMax);
                 expect(result).toBe(
                     COMMON_TEXT_ADMIN.VALIDATION_MESSAGE.getMaxError(
@@ -162,10 +164,9 @@ describe('reports-media-settings-schema', () => {
                 );
             });
 
-            it('should return undefined for value at max', () => {
-                const result = REPORTS_CHANGED_LIVES_VALIDATION_FUNCTIONS.validateChangedLives(
-                    REPORTS_MEDIA_SETTINGS_CHANGED_LIVES_VALIDATION.changedLives.max,
-                );
+            it('should return undefined for value length at max', () => {
+                const maxVal = Number('1'.repeat(REPORTS_MEDIA_SETTINGS_CHANGED_LIVES_VALIDATION.changedLives.max));
+                const result = REPORTS_CHANGED_LIVES_VALIDATION_FUNCTIONS.validateChangedLives(maxVal);
                 expect(result).toBeUndefined();
             });
         });
--- a/src/validation/admin/reports-schema/reports-media-settings/reports-media-settings-schema.test.ts
+++ b/src/validation/admin/reports-schema/reports-media-settings/reports-media-settings-schema.test.ts
@@ -152,10 +152,8 @@ describe('reports-media-settings-schema', () => {
                 expect(result).toBe(REPORTS_TEXT.MESSAGE.INVALID_VALUE);
             });
 
-            it('should return max error when value length exceeds max', () => {
-                const overMax = Number(
-                    '1'.repeat(REPORTS_MEDIA_SETTINGS_CHANGED_LIVES_VALIDATION.changedLives.max + 1),
-                );
+            it('should return max error when value exceeds max numeric bound', () => {
+                const overMax = 10 ** REPORTS_MEDIA_SETTINGS_CHANGED_LIVES_VALIDATION.changedLives.max;
                 const result = REPORTS_CHANGED_LIVES_VALIDATION_FUNCTIONS.validateChangedLives(overMax);
                 expect(result).toBe(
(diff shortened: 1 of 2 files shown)


src/validation/admin/reports-schema/reports-media-settings/reports-media-settings-schema.ts
```diff
@@ -61,11 +61,15 @@ const changedLivesSchema = Yup.object({
...
-            REPORTS_MEDIA_SETTINGS_CHANGED_LIVES_VALIDATION.changedLives.max,
+        .test(
+            'max-length',
             COMMON_TEXT_ADMIN.VALIDATION_MESSAGE.getMaxError(
                 REPORTS_MEDIA_SETTINGS_CHANGED_LIVES_VALIDATION.changedLives.max,
             ),
+            (value) =>
+                value === undefined ||
+                value === null ||
+                String(value).length <= REPORTS_MEDIA_SETTINGS_CHANGED_LIVES_VALIDATION.changedLives.max,
```
mehalyna: The validation uses `<=` (less than or equal), but the constant is named `max` with value `10`. This means it will accept 10-character numbers, but the error message says "Не більше 10 символів" (No more than 10 characters), which is ambiguous.

**Problem scenarios:**
- `1234567890` (10 digits) - **Passes**
- `12345678901` (11 digits) - **Fails** 

This is **inconsistent with the typical interpretation** of "max 10 characters" which usually means **up to and including 10**, but the field name suggests it should be an **upper bound**. 

Does "max 10 characters" mean:
- A) Accepts up to 10 characters (current behavior with `<=`)
- B) Accepts fewer than 10 characters (should use `<`)

If the business rule is "must be fewer than 10 characters", then this is **functionally incorrect**.

mehalyna: The schema has `.required()` but the custom test only checks `undefined` and `null`, not empty string.

*Test case:*
```
String('').length  // 0, which is <= 10, so it passes the length check
```

However, Yup's `.required()` should catch this **before** the custom test runs, BUT:

```
.integer(REPORTS_TEXT.MESSAGE.INVALID_VALUE)
```

The `.integer()` constraint comes **before** the length test. Empty string is **not a valid integer**, so it will fail at `.integer()` with "Invalid value" error, not at the length check.

Not a critical issue because `.integer()` will catch it, but the custom test is redundant for empty string since `.required()` handles it.

mehalyna: Negative numbers include a `-` character in their string representation.

*Test case:*
```
String(-123456789).length  // 10 characters (including the minus sign)
String(-1234567890).length // 11 characters - Would fail length check
```

However, the schema has:
```
.min(2, COMMON_TEXT_ADMIN.VALIDATION_MESSAGE.getMinError(2))
```

This `.min(2)` constraint prevents negative numbers **as numeric values**, so `-5` would fail the minimum check.

Not a critical issue because `.min(2)` prevents negative numbers, but the validation semantics are confusing (mixing numeric constraints with string length constraints).

src/validation/admin/reports-schema/reports-media-settings/reports-media-settings-schema.test.ts
```diff
@@ -162,10 +164,9 @@ describe('reports-media-settings-schema', () => {
                 );
             });
 
-            it('should return undefined for value at max', () => {
-                const result = REPORTS_CHANGED_LIVES_VALIDATION_FUNCTIONS.validateChangedLives(
-                    REPORTS_MEDIA_SETTINGS_CHANGED_LIVES_VALIDATION.changedLives.max,
-                );
+            it('should return undefined for value length at max', () => {
+                const maxVal = Number('1'.repeat(REPORTS_MEDIA_SETTINGS_CHANGED_LIVES_VALIDATION.changedLives.max));
+                const result = REPORTS_CHANGED_LIVES_VALIDATION_FUNCTIONS.validateChangedLives(maxVal);
```
mehalyna: The test creates a 10-digit number (`1111111111`) and expects it to **pass**. This confirms the `<=` behavior.

What about `9999999999` (max 10-digit value)? The test uses all `1`s, which doesn't test numeric overflow or boundary edge cases.

*Also missing* Test for exactly 9 characters, to verify the boundary below the maximum.

Not critical, but test coverage could be improved.

src/validation/admin/reports-schema/reports-media-settings/reports-media-settings-schema.ts
mehalyna: This converts the number to a string on **every keystroke** during form input. For a single field, the performance impact is negligible, but it's worth noting that this is less efficient than numeric comparison.

Not a critical issue, but architecturally, validating character length for a numeric field is semantically odd.

mehalyna (approved):

PR #480, opened 2026-07-07
Fix/issue 2765 [Partners] [Admin] The "Опублікувати" button is enabled without pending changes

fhyhtyh:
## Github ticket

* [Github ticket](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2765)

## Description

This PR resolves an issue in the Admin Panel's Partners page where the "Опублікувати" (Publish) buttons for the main Banner (Title) and individual Partner sections were incorrectly enabled by default. The underlying form logic has been updated to implement comprehensive "dirty-state" tracking. Now, the publish buttons remain disabled upon loading and will only become active when the user makes actual modifications to the form fields (e.g., editing text, changing images, or adding/removing partners).

## How it looks

https://github.com/user-attachments/assets/a6e616de-6b93-4e26-9a4f-1a783e58c987

## Summary of change

*   **Dirty State Tracking for Banner (`PartnerBannerForm.tsx`)**: Added a `savedValues` state and an `isDirty` calculation hook to compare current form values against the originally loaded data, disabling the publish button when unmodified.
*   **Deep Comparison for Sections (`PartnerSectionsEditor.tsx`)**: Implemented the `isSectionDirty` helper function to perform a deep comparison between local section forms (including nested partner data) and the original `fetchedSections` API response.
*   **Component Prop Updates (`PartnerSectionForm.tsx`)**: Extended `PartnerSectionProps` to accept an `isDirty` boolean, allowing the component to conditionally disable its publish button.
*   **Test Suite Updates**: Modified unit tests in `PartnerBannerForm.test.tsx` and `PartnerSectionForm.test.tsx` to simulate user changes prior to publishing, successfully testing the new disabled-by-default behavior.

## How to recreate changes

1. Log into the Admin panel and navigate to the **"Партнери"** page.
2. Locate the **"Опублікувати"** (Publish) buttons for both the main Title/Banner section and any individual Partner sections. Observe that they are disabled by default.
3. Make a change to any field (e.g., add a character to a description, change an image, or add a new partner).
4. Observe that the corresponding **"Опублікувати"** button immediately becomes enabled.
5. Click the **"Опублікувати"** button and confirm the publication in the popup modal.
6. Verify that after the success toast appears, the **"Опублікувати"** button immediately becomes disabled again until further changes are made.

## CHECK LIST

## Summary by CodeRabbit

* **New Features**
  * Publish actions in partner banners and sections now require actual changes before becoming available.
  * The editor now better reflects whether a section has unsaved changes.

* **Bug Fixes**
  * Improved publish behavior so saved content is tracked correctly after successful updates.
  * Deleting a section now keeps the displayed list in sync with the latest state.
  * Updated form handling to ensure publish button states match validation and change status.

### DIFF
--- a/src/pages/admin/partners/components/partner-banner-form/PartnerBannerForm.tsx
+++ b/src/pages/admin/partners/components/partner-banner-form/PartnerBannerForm.tsx
@@ -50,11 +50,26 @@ const isFormValid = (values: PartnerBannerValues, errors: PartnerBannerErrorStat
     return !titleError && !descError && !errors.image;
 };
 
+interface FormState {
+    values: PartnerBannerValues | null;
+    savedValues: PartnerBannerValues | null;
+}
+
+const isImageEqual = (img1: ImageValues | Image | null, img2: ImageValues | Image | null) => {
+    if (img1 === img2) return true;
+    if (!img1 || !img2) return false;
+    if ('url' in img1 && 'url' in img2) return img1.url === img2.url;
+    if ('base64' in img1 && 'base64' in img2) return img1.base64 === img2.base64;
+    return false;
+};
+
 export const PartnerBanner = () => {
     const client = useAdminClient();
     const { addToast } = useToast();
-    const [values, setValues] = useState<PartnerBannerValues | null>(null);
-    const [savedValues, setSavedValues] = useState<PartnerBannerValues | null>(null);
+    const [formState, setFormState] = useState<FormState>({
+        values: null,
+        savedValues: null,
+    });
     const [errors, setErrors] = useState<PartnerBannerErrorState>({});
     const [touched, setTouched] = useState<{ title?: boolean; description?: boolean }>({});
     const [isPublishing, setIsPublishing] = useState(false);
@@ -76,20 +91,21 @@ export const PartnerBanner = () => {
         autoFetchDisabled: false,
     });
 
+    const { values, savedValues } = formState;
+
     const isDirty = React.useMemo(() => {
         if (!savedValues || !values) return false;
         return (
             savedValues.title !== values.title ||
             savedValues.description !== values.description ||
             savedValues.imageId !== values.imageId ||
-            savedValues.image !== values.image
+            !isImageEqual(savedValues.image, values.image)
         );
     }, [savedValues, values]);
 
     useEffect(() => {
         if (!isLoadingData && bannerData && bannerData.title) {
-            setValues(bannerData);
-            setSavedValues(bannerData);
+            setFormState({ values: bannerData, savedValues: bannerData });
             setTouched({});
             setErrors({});
             setTitleKey((prev) => prev + 1);
@@ -110,7 +126,10 @@ export const PartnerBanner = () => {
     }, [fetchError, addToast]);
 
     const handleTitleChange = useCallback((newValue: string) => {
-        setValues((prev) => (prev ? { ...prev, title: newValue } : null));
+        setFormState((prev) => ({
+            ...prev,
+            values: prev.values ? { ...prev.values, title: newValue } : null,
+        }));
 
         const plainText = getPlainTextFromHtml(newValue);
 
@@ -126,7 +145,10 @@ export const PartnerBanner = () => {
 
     const handleDescriptionChange = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
         const newValue = e.target.value;
-        setValues((prev) => (prev ? { ...prev, description: newValue } : null));
+        setFormState((prev) => ({
+            ...prev,
+            values: prev.values ? { ...prev.values, description: newValue } : null,
+        }));
         setTouched((prev) => ({ ...prev, description: true }));
         setErrors((prev) => ({
             ...prev,
@@ -135,7 +157,10 @@ export const PartnerBanner = () => {
     }, []);
 
     const handleImageChange = useCallback((value: ImageValues | null) => {
-        setValues((prev) => (prev ? { ...prev, image: value, imageId: value ? prev.imageId : null } : null));
+        setFormState((prev) => ({
+            ...prev,
+            values: prev.values ? { ...prev.values, image: value, imageId: value ? prev.values.imageId : null } : null,
+        }));
     }, []);
 
     const handleImageError = useCallback((error: string | null) => {
@@ -162,8 +187,7 @@ export const PartnerBanner = () => {
                 imageId: values.imageId,
             });
 
-            setValues(updatedBanner);
-            setSavedValues(updatedBanner);
+            setFormState({ values: updatedBanner, savedValues: updatedBanner });
             setTouched({});
             setErrors({});
             addToast(PARTNERS_TEXT.MESSAGE.BANNER_SAVED, ToastType.Success);
--- a/src/pages/admin/partners/components/partner-banner-form/PartnerBannerForm.tsx
+++ b/src/pages/admin/partners/components/partner-banner-form/PartnerBannerForm.tsx
@@ -54,6 +54,7 @@ export const PartnerBanner = () => {
     const client = useAdminClient();
     const { addToast } = useToast();
     const [values, setValues] = useState<PartnerBannerValues | null>(null);
+    const [savedValues, setSavedValues] = useState<PartnerBannerValues | null>(null);
     const [errors, setErrors] = useState<PartnerBannerErrorState>({});
     const [touched, setTouched] = useState<{ title?: boolean; description?: boolean }>({});
     const [isPublishing, setIsPublishing] = useState(false);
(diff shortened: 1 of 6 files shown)


src/pages/admin/partners/components/partner-banner-form/PartnerBannerForm.tsx
```diff
@@ -75,9 +76,20 @@ export const PartnerBanner = () => {
...
+            savedValues.description !== values.description ||
+            savedValues.imageId !== values.imageId ||
+            savedValues.image !== values.image
+        );
+    }, [savedValues, values]);
+
     useEffect(() => {
         if (!isLoadingData && bannerData && bannerData.title) {
             setValues(bannerData);
+            setSavedValues(bannerData);
```
mehalyna: Two separate `setState` calls are not atomic. React may batch them, but in edge cases (especially with Concurrent Mode or Suspense), the component could render between these calls, causing:
- `isDirty` to incorrectly return `true` momentarily
- Publish button to flash enabled/disabled

*Better approach:*
```
useEffect(() => {
    if (!isLoadingData && bannerData && bannerData.title) {
        const data = bannerData;
        setValues((prev) => data);
        setSavedValues((prev) => data);
        // OR combine into single state:
        // setState({ values: data, savedValues: data })
    }
}, [isLoadingData, bannerData]);
```

src/pages/admin/partners/components/partner-banner-form/PartnerBannerForm.tsx
```diff
@@ -75,9 +76,20 @@ export const PartnerBanner = () => {
...
 
+    const isDirty = React.useMemo(() => {
+        if (!savedValues || !values) return false;
+        return (
+            savedValues.title !== values.title ||
+            savedValues.description !== values.description ||
+            savedValues.imageId !== values.imageId ||
+            savedValues.image !== values.image
+        );
+    }, [savedValues, values]);
```
mehalyna, line 104: The `image` field is likely an **object** (e.g., `{ url: string, id: number }`), and this comparison uses `!==` which checks **reference equality**, not deep equality.

*Problem scenarios:*
1. User uploads same image again - Different object reference - `isDirty = true`  (correct)
2. Form re-renders with same image data but new object - `isDirty = true`  (false positive)

*Evidence from the code:*
```
// Line 54: values is typed as PartnerBannerValues | null
// PartnerBannerValues likely has structure:
// { title: string, description: string, imageId: number, image: ImageDto }
```

If `image` is a primitive (like a string URL), this is fine. But if it's an object, this will cause false positives.

**Fix:**
```
savedValues.imageId !== values.imageId ||
savedValues.image?.url !== values.image?.url // Deep comparison if image is object
```

Or use a deep equality library:
```
import { isEqual } from 'lodash';

const isDirty = React.useMemo(() => {
    if (!savedValues || !values) return false;
    return !isEqual(savedValues, values);
}, [savedValues, values]);
```

src/pages/admin/partners/components/partner-sections-editor/PartnerSectionsEditor.tsx
```diff
@@ -38,6 +38,31 @@ const isSectionEmpty = (section: PartnerSectionFormValues): boolean => {
...
+        const originalPartner = originalSection.partners.find((p) => p.id === partner.partnerId);
+        if (!originalPartner) return true;
+
+        if (partner.description !== originalPartner.description) return true;
+        if (partner.imageId !== originalPartner.imageId) return true;
+        if (partner.image !== originalPartner.image) return true;
+    }
+
+    return false;
+};
```
mehalyna, line 74: For each partner in the section, it searches through all original partners. With large partner lists (e.g., 100+ partners), this could cause noticeable lag on every render.

**Fix:** Use a Map for O(1) lookup:
```
const originalPartnersMap = new Map(originalSection.partners.map(p => [p.id, p]));

for (const partner of section.partners) {
    if (partner.partnerId === null) return true;
    
    const originalPartner = originalPartnersMap.get(partner.partnerId);
    if (!originalPartner) return true;
    // ...
}
```

src/pages/admin/partners/components/partner-sections-editor/PartnerSectionsEditor.tsx
```diff
@@ -331,6 +371,7 @@ export const PartnerSectionsEditor = forwardRef<PartnerSectionsEditorRef>((_, re
                     value={section}
                     errors={errors[index] || { partners: [] }}
                     disabled={isSectionsLoading || isPublishing}
+                    isDirty={isSectionDirty(section, fetchedSections)}
```
mehalyna: This function is called **inline during render** for every section, on every render. It performs deep comparisons without memoization.

*Problem:*
- If there are 10 sections, this runs 10 times per render
- Each call does O(n²) partner comparison
- Triggers on every state change in the parent component

**Fix:** Memoize the result:
```
const sectionDirtyStates = React.useMemo(
    () => localSections.map(section => isSectionDirty(section, fetchedSections)),
    [localSections, fetchedSections]
);

// Then in render:
isDirty={sectionDirtyStates[index]}
```

src/pages/admin/partners/components/partner-sections-editor/PartnerSectionsEditor.tsx
```diff
@@ -216,7 +252,7 @@ export const PartnerSectionsEditor = forwardRef<PartnerSectionsEditorRef>((_, re
             setIsPublishing(false);
             setSectionToPublishId(null);
         }
-    }, [localSections, addToast, client, setLocalSections, sectionToPublishId]);
+    }, [localSections, addToast, client, setLocalSections, sectionToPublishId, setFetchedSections]);
```
mehalyna: The `handlePublish` callback depends on `localSections`, but the dependency might be stale if the callback is triggered during a state update.

*Problem scenario:*
1. User clicks publish
2. During API call, user makes another change to a different section
3. `localSections` in the callback closure is now stale
4. When `setFetchedSections` runs, it might use outdated section data

**Fix:** Use functional state updates:
```
setFetchedSections((current) => {
    const currentArray = current || [];
    const section = localSections.find(s => s.localId === sectionToPublishId);
    // ^^^ This 'localSections' is from closure, might be stale
```

Should be:
```
setFetchedSections((current) => {
    // Use 'current' state or pass section data as parameter
});
```

**Better pattern:** Pass the published section data explicitly:
```
const handlePublishRequest = useCallback((localId: string, sectionData: PartnerSectionFormValues) => {
    setSectionToPublishId(localId);
    setSectionToPublish(sectionData); // Store the data
});
```

src/pages/admin/partners/components/partner-sections-editor/PartnerSectionsEditor.tsx
```diff
@@ -38,6 +38,31 @@ const isSectionEmpty = (section: PartnerSectionFormValues): boolean => {
...
+    if (section.partners.length !== originalSection.partners.length) return true;
+
+    for (const partner of section.partners) {
+        if (partner.partnerId === null) return true;
+
+        const originalPartner = originalSection.partners.find((p) => p.id === partner.partnerId);
+        if (!originalPartner) return true;
+
+        if (partner.description !== originalPartner.description) return true;
+        if (partner.imageId !== originalPartner.imageId) return true;
```
mehalyna, line 69: Same problem  - `partner.image !== originalPartner.image` compares object references, not content.

*Scenario:*
1. Component re-renders
2. Partner image object is reconstructed with same data but different reference
3. `isSectionDirty` returns `true` even though nothing changed
4. Publish button incorrectly enabled

**Fix:**
```
if (partner.imageId !== originalPartner.imageId) return true;
// Remove the image object comparison since imageId is sufficient
```

Or deep compare:
```
if (partner.image?.url !== originalPartner.image?.url) return true;
```

src/pages/admin/partners/components/partner-section/PartnerSectionForm.test.tsx
mehalyna, line 1: All tests pass `isDirty={true}`, which means:
- Tests don't verify the button is disabled when `isDirty={false}`
- Tests don't verify the dirty state calculation logic
- False sense of security - tests would pass even if dirty logic is completely broken

**Required test cases:**
```
it('disables publish button when not dirty', () => {
    render(<PartnerSectionForm {...props} isDirty={false} />);
    expect(getPublishButton()).toBeDisabled();
});

it('enables publish button only when dirty and valid', () => {
    const { rerender } = render(<PartnerSectionForm {...props} isDirty={false} />);
    expect(getPublishButton()).toBeDisabled();
    
    rerender(<PartnerSectionForm {...props} isDirty={true} />);
    expect(getPublishButton()).toBeEnabled();
});
```

Test coverage is misleading. Bugs in dirty state tracking won't be caught.

mehalyna (approved):

PR #484, opened 2026-07-08
Feature/issue 2617 [Main] Disable the ability to translate

fhyhtyh:
## Github ticket

* [Github ticket](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2617)

## Description

This PR implements logic to prevent inconsistent translations on the Admin Main Page by disabling the translation icon when a block contains unpublished changes in its text input fields. It introduces an dirty-field checking mechanism that accurately compares current rich text field inputs against saved data to guarantee translations are only performed on published content.

## How it looks
image

## Summary of change

- **State Management (`MainPageContent.tsx`)**: Added the `BLOCK_TEXT_FIELDS` mapping to precisely identify and track the translatable text input fields for each specific block.
- **Dirty Checking Logic**: Introduced the `hasBlockDirtyTextFields` helper function to evaluate if any text fields within the active block differ from their published state, utilizing the existing rich text HTML normalization (`areRichTextFieldValuesEqual`) to avoid false positives.
- **UI Interaction**: Bound the `disabled` property of the translation `IconButton` to the new dirty state logic. Updated `handleOpenTranslationModal` to exit early to cover edge cases.
- **Testing (`MainPageContent.test.tsx`)**: Added a new test case to assert that the translation icon correctly becomes disabled when modifying text input fields in a translatable block.

## How to recreate changes

1. Log in as an Admin and navigate to the **Головна** page.
2. Select any tab with translatable content (e.g., Title, About Us, Donations, Partners).
3. Observe that the translation icon is visible and enabled in the top right corner of the content area.
4. Modify any text input field within the active block (e.g., type into the title or description field).
5. Observe that the translation icon immediately becomes disabled, preventing you from adding a translation to unpublished content.
6. Either revert the text back to its original state or save the changes by clicking **"Опублікувати" (Publish)**.
7. Observe that the translation icon becomes enabled again once the state is consistent.

## CHECK LIST

## Summary by CodeRabbit

* **Bug Fixes**
  * The translation action is now disabled only when the currently selected content block has unsaved text changes, preventing translation while edits are pending.
  * The translation action is re-enabled after the dirty text is cleaned back to an equal/rich-text value.

* **Tests**
  * Updated and expanded coverage to verify the translation icon’s enabled/disabled state across dirty and clean text transitions using UI state assertions.

### DIFF
--- a/src/pages/admin/main/components/main-page-content/MainPageContent.test.tsx
+++ b/src/pages/admin/main/components/main-page-content/MainPageContent.test.tsx
@@ -1263,4 +1263,19 @@ describe('MainPageContent', () => {
 
         await waitFor(() => expect(screen.queryByText('Додати переклад')).not.toBeInTheDocument());
     });
+
+    it('disables the translation icon when text input fields have unpublished changes', async () => {
+        mockLocalizationToolkitState.translationLanguages = [{ id: 2, code: 'en', name: 'Англійська' }];
+
+        await renderAndLoadContent();
+
+        const translateIcon = screen.getByLabelText('Додати переклад');
+        expect(translateIcon).not.toBeDisabled();
+
+        fireEvent.click(screen.getByTestId('publish-btn-dirty'));
+
+        await waitFor(() => {
+            expect(translateIcon).toBeDisabled();
+        });
+    });
 });
--- a/src/pages/admin/main/components/main-page-content/MainPageContent.test.tsx
+++ b/src/pages/admin/main/components/main-page-content/MainPageContent.test.tsx
@@ -114,6 +114,14 @@ const MockFormBlock = ({ testId, btnTestId, onPublish, isPublishDisabled, isRead
             >
                 Normalize Dirty
             </button>
+            <button
+                data-testid={`${btnTestId}-clean-text`}
+                onClick={() => {
+                    setValue('titleUa', 'Test Title', { shouldDirty: true, shouldValidate: true });
+                }}
+            >
+                Clean Text
+            </button>
             <button data-testid={btnTestId} onClick={onPublish} disabled={isPublishDisabled}>
                 Publish
             </button>
@@ -1276,6 +1284,24 @@ describe('MainPageContent', () => {
 
         await waitFor(() => {
             expect(translateIcon).toBeDisabled();
+        });
+    });
+
+    it('re-enables translation icon when changes are reverted', async () => {
+        mockLocalizationToolkitState.translationLanguages = [{ id: 2, code: 'en', name: 'Англійська' }];
+
+        await renderAndLoadContent();
+
+        const translateIcon = screen.getByLabelText('Додати переклад');
+
+        fireEvent.click(screen.getByTestId('publish-btn-dirty'));
+        await waitFor(() => {
+            expect(translateIcon).toBeDisabled();
+        });
+
+        fireEvent.click(screen.getByTestId('publish-btn-clean-text'));
+        await waitFor(() => {
+            expect(translateIcon).not.toBeDisabled();
         });
     });
 });
--- a/src/pages/admin/main/components/main-page-content/MainPageContent.tsx
+++ b/src/pages/admin/main/components/main-page-content/MainPageContent.tsx
@@ -80,13 +80,6 @@ const TABS: TabItem[] = [
     { id: 'partners', label: MAIN_PAGE_TEXT.TABS.PARTNERS, localizationBlock: MainPageLocalizationBlock.Partners },
 ];
 
-const TRANSLATABLE_BLOCKS = new Set<MainPageLocalizationBlock>([
-    MainPageLocalizationBlock.Title,
-    MainPageLocalizationBlock.AboutUs,
-    MainPageLocalizationBlock.Donations,
-    MainPageLocalizationBlock.Partners,
-]);
-
 const sanitizeMainPageFormValues = (values: MainPageFormValues): MainPageFormValues => ({
     ...values,
     statisticsTitleUa: values.statisticsTitleUa ?? '',
@@ -126,7 +119,16 @@ const areRichTextFieldValuesEqual = (
     normalizeRichTextHtmlForComparison(String(currentValues[field] ?? '')) ===
     normalizeRichTextHtmlForComparison(String(savedValues[field] ?? ''));
 
-const BLOCK_TEXT_FIELDS: Partial<Record<MainPageLocalizationBlock, Array<keyof MainPageFormValues>>> = {
+const TRANSLATABLE_BLOCKS_ARRAY = [
+    MainPageLocalizationBlock.Title,
+    MainPageLocalizationBlock.AboutUs,
+    MainPageLocalizationBlock.Donations,
+    MainPageLocalizationBlock.Partners,
+] as const;
+
+type TranslatableBlock = (typeof TRANSLATABLE_BLOCKS_ARRAY)[number];
+
+const BLOCK_TEXT_FIELDS: Record<TranslatableBlock, Array<keyof MainPageFormValues>> = {
     [MainPageLocalizationBlock.Title]: ['titleUa', 'titleEn', 'descriptionUa', 'descriptionEn'],
     [MainPageLocalizationBlock.AboutUs]: [
         'aboutUsTitleUa',
@@ -148,6 +150,8 @@ const BLOCK_TEXT_FIELDS: Partial<Record<MainPageLocalizationBlock, Array<keyof M
     ],
 };
 
+const TRANSLATABLE_BLOCKS = new Set<MainPageLocalizationBlock>(TRANSLATABLE_BLOCKS_ARRAY);
+
 const hasBlockDirtyTextFields = (
     block: MainPageLocalizationBlock | undefined,
     dirtyFields: Partial<Record<keyof MainPageFormValues, unknown>>,
@@ -156,7 +160,7 @@ const hasBlockDirtyTextFields = (
 ) => {
     if (block == null) return false;
 
-    const blockFields = BLOCK_TEXT_FIELDS[block];
+    const blockFields = BLOCK_TEXT_FIELDS[block as TranslatableBlock];
     if (!blockFields) return false;
 
     return blockFields.some((fieldName) => {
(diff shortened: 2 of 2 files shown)


src/pages/admin/main/components/main-page-content/MainPageContent.tsx
```diff
@@ -464,8 +508,18 @@ export const MainPageContent = () => {
...
         TRANSLATABLE_BLOCKS.has(selectedTab.localizationBlock);
 
+    const isTranslateButtonDisabled =
+        canTranslateActiveBlock &&
+        hasBlockDirtyTextFields(
+            selectedTab.localizationBlock,
+            methods.formState.dirtyFields,
+            methods.getValues(),
+            savedValuesRef.current,
+        );
```
mehalyna, line 514: `methods.getValues()` is called on **every render** to check if translate button should be disabled. This:
- Calls `hasBlockDirtyTextFields` → iterates through all block fields
- For each field, calls `areRichTextFieldValuesEqual` → normalizes HTML strings
- Runs regex replacements and string comparisons

With 4 rich text fields per block, this means 4+ regex operations on every keystroke.

*Suggestion:*
```
const isTranslateButtonDisabled = useMemo(
    () =>
        canTranslateActiveBlock &&
        hasBlockDirtyTextFields(
            selectedTab.localizationBlock,
            methods.formState.dirtyFields,
            methods.getValues(),
            savedValuesRef.current,
        ),
    [
        canTranslateActiveBlock,
        selectedTab.localizationBlock,
        methods.formState.dirtyFields,
        // Don't include getValues() - it returns new object each time
        // Use watch() or formState instead
    ]
);
```

Better approach - use form state directly:
```
const dirtyTextFields = useMemo(() => {
    if (!selectedTab.localizationBlock) return false;
    const blockFields = BLOCK_TEXT_FIELDS[selectedTab.localizationBlock];
    if (!blockFields) return false;
    
    return blockFields.some(field => methods.formState.dirtyFields[field]);
}, [
    selectedTab.localizationBlock,
    methods.formState.dirtyFields,
]);

const isTranslateButtonDisabled = canTranslateActiveBlock && dirtyTextFields;
```

src/pages/admin/main/components/main-page-content/MainPageContent.tsx
```diff
@@ -464,8 +508,18 @@ export const MainPageContent = () => {
         selectedTab.localizationBlock != null &&
         TRANSLATABLE_BLOCKS.has(selectedTab.localizationBlock);
 
+    const isTranslateButtonDisabled =
+        canTranslateActiveBlock &&
+        hasBlockDirtyTextFields(
+            selectedTab.localizationBlock,
+            methods.formState.dirtyFields,
+            methods.getValues(),
```
mehalyna, line 512: `getValues()` returns the **current form state**, but `react-hook-form` may not have committed the latest keystroke yet. This means:
- User types "Hello" → validation runs with "Hell"
- Rich text normalization might return false positive/negative

*Scenario:*
1. Saved: `<p>Hello</p>`
2. User types → Current: `<p>Hello!</p>`
3. `getValues()` might return `<p>Hello</p>` (not yet committed)
4. Button stays enabled when it should be disabled

**Suggestion:*
```
// Use watch() for reactive updates
const watchedValues = methods.watch();

const isTranslateButtonDisabled = useMemo(
    () =>
        canTranslateActiveBlock &&
        hasBlockDirtyTextFields(
            selectedTab.localizationBlock,
            methods.formState.dirtyFields,
            watchedValues,  // ← Always latest
            savedValuesRef.current,
        ),
    [canTranslateActiveBlock, selectedTab.localizationBlock, methods.formState.dirtyFields, watchedValues]
);
```

src/pages/admin/main/components/main-page-content/MainPageContent.tsx
```diff
@@ -126,6 +126,50 @@ const areRichTextFieldValuesEqual = (
...
+        'donationsDescriptionUa',
+        'donationsDescriptionEn',
+    ],
+    [MainPageLocalizationBlock.Partners]: [
+        'partnersTitleUa',
+        'partnersTitleEn',
+        'partnersDescriptionUa',
+        'partnersDescriptionEn',
+    ],
+};
```
mehalyna: Using `Partial<Record<...>>` means some blocks might be missing. If a new translatable block is added but not included in `BLOCK_TEXT_FIELDS`, translation won't be disabled for that block's dirty fields.

*Suggestion:*u
```
// Use Required instead of Partial for translatable blocks
const TRANSLATABLE_BLOCK_TEXT_FIELDS: Record<
    Extract<
        MainPageLocalizationBlock,
        MainPageLocalizationBlock.Title | 
        MainPageLocalizationBlock.AboutUs | 
        MainPageLocalizationBlock.Donations | 
        MainPageLocalizationBlock.Partners
    >,
    Array<keyof MainPageFormValues>
> = {
    [MainPageLocalizationBlock.Title]: ['titleUa', 'titleEn', 'descriptionUa', 'descriptionEn'],
    [MainPageLocalizationBlock.AboutUs]: [/* ... */],
    [MainPageLocalizationBlock.Donations]: [/* ... */],
    [MainPageLocalizationBlock.Partners]: [/* ... */],
    // TypeScript will error if any translatable block is missing
};

const hasBlockDirtyTextFields = (
    block: MainPageLocalizationBlock | undefined,
    // ...
) => {
    if (block == null) return false;
    
    const blockFields = TRANSLATABLE_BLOCK_TEXT_FIELDS[block as keyof typeof TRANSLATABLE_BLOCK_TEXT_FIELDS];
    if (!blockFields) return false;  // Non-translatable block
```

src/pages/admin/main/components/main-page-content/MainPageContent.tsx
```diff
@@ -126,6 +126,50 @@ const areRichTextFieldValuesEqual = (
...
+        if (!dirtyFields[fieldName]) return false;
+
+        if (RICH_TEXT_FORM_FIELD_SET.has(fieldName)) {
+            return !areRichTextFieldValuesEqual(fieldName, currentValues, savedValues);
+        }
+
+        return currentValues[fieldName] !== savedValues[fieldName];
+    });
+};
+
```
mehalyna, line 172: `dirtyFields` is controlled by `react-hook-form` and only resets when form is explicitly reset. If user:
1. Types "Modified text" → `dirtyFields.titleUa = true`
2. Manually deletes and types back original → `currentValues.titleUa === savedValues.titleUa`
3. But `dirtyFields.titleUa` **still true**

Result: Translation stays disabled even though content matches saved state.

*Suggestion:*
```
return blockFields.some((fieldName) => {
    // Don't rely solely on dirtyFields
    if (RICH_TEXT_FORM_FIELD_SET.has(fieldName)) {
        return !areRichTextFieldValuesEqual(fieldName, currentValues, savedValues);
    }
    return currentValues[fieldName] !== savedValues[fieldName];
});
// Remove the dirtyFields check - just compare values directly
```

Or keep dirtyFields for performance but add value comparison:
```typescript
return blockFields.some((fieldName) => {
    if (!dirtyFields[fieldName]) return false;
    
    if (RICH_TEXT_FORM_FIELD_SET.has(fieldName)) {
        const isEqual = areRichTextFieldValuesEqual(fieldName, currentValues, savedValues);
        if (isEqual) {
            // Field is marked dirty but content matches saved - clear dirty flag
            methods.formState.dirtyFields[fieldName] = undefined;
            return false;
        }
        return true;
    }
    
    const currentVal = currentValues[fieldName];
    const savedVal = savedValues[fieldName];
    return currentVal !== savedVal;
});
```

src/pages/admin/main/components/main-page-content/MainPageContent.test.tsx
mehalyna, line 1: Test only checks that button becomes disabled after change. Missing test case:
- User makes change → Button disabled ✓
- User reverts change → Button should re-enable (not tested)

**Suggestion:**
```typescript
it('re-enables translation icon when changes are reverted', async () => {
    await renderAndLoadContent();
    
    const translateIcon = screen.getByLabelText('Додати переклад');
    const inputField = screen.getByTestId('title-input');
    const originalValue = inputField.value;
    
    // Make change
    fireEvent.change(inputField, { target: { value: 'Modified' } });
    await waitFor(() => expect(translateIcon).toBeDisabled());
    
    // Revert change
    fireEvent.change(inputField, { target: { value: originalValue } });
    await waitFor(() => expect(translateIcon).not.toBeDisabled());
});
```

src/pages/admin/main/components/main-page-content/MainPageContent.tsx
```diff
@@ -464,8 +508,18 @@ export const MainPageContent = () => {
...
+        hasBlockDirtyTextFields(
+            selectedTab.localizationBlock,
+            methods.formState.dirtyFields,
+            methods.getValues(),
+            savedValuesRef.current,
+        );
+
     const handleOpenTranslationModal = () => {
         if (
+            isTranslateButtonDisabled ||
```
mehalyna: The button is already `disabled={isTranslateButtonDisabled}`, so `onClick` won't fire when disabled. The check is redundant.

Good for defensive programming in case button is triggered programmatically, but current logic makes `isTranslateButtonDisabled` less comprehensive than the full check.

*Suggestion:*

Either remove the redundant check:
```
const handleOpenTranslationModal = () => {
    // Button is already disabled, no need to check again
    if (
        translationLanguages.length === 0 ||
        selectedTab.localizationBlock == null ||
        !TRANSLATABLE_BLOCKS.has(selectedTab.localizationBlock)
    ) {
        return;
    }
```

Or make `isTranslateButtonDisabled` more comprehensive:
```
const isTranslateButtonDisabled = 
    !canTranslateActiveBlock ||
    translationLanguages.length === 0 ||
    hasBlockDirtyTextFields(...);
```

src/pages/admin/main/components/main-page-content/MainPageContent.tsx
fhyhtyh: RHF optimizes performance by wrapping formState in a Proxy over a mutable reference. Because the object reference doesn't consistently change when internal values update, React's useMemo fails to detect changes. The original function contains if (!dirtyFields[fieldName]) return false;. This means the heavy regex is never executed for unmodified fields.

src/pages/admin/main/components/main-page-content/MainPageContent.tsx
fhyhtyh: getValues() is a synchronous read from RHF's internal store. It is never "stale" relative to the current render cycle. Using watch() makes the component re-render on every keystroke which can cause performance issues.

src/pages/admin/main/components/main-page-content/MainPageContent.tsx
fhyhtyh: React Hook Form dynamically tracks dirtyFields by continuously running a deep comparison between the current values and the defaultValues. When you manually delete the text and type back the original, RHF sees that the current value matches the defaultValues again, and it automatically removes that field from dirtyFields.

mehalyna (approved):

PR #488, opened 2026-07-11
Fix/issue-1834 [Reports] Edit the sum of expenses of a program allocated

yaroslavkucher:
## JIRA or Github ticker

* [1834](https://github.com/ita-social-projects/VictoryCenter-Back/issues/1834)

## Description

This PR implements inline editing functionality for program expenses records.
As an Admin, I can now edit the UAH and USD amounts directly within the "Програмні витрати" (Program expenses) tab. This ensures that any mistakes in record details can be quickly fixed to present relevant information to users, matching the behavior previously implemented for funds and expenditures. 

## How it looks

image

## Summary of change

* **ProgramExpensesSection/ReportAnalytics:** Refactored `ProgramExpensesSection` to manage its own active state by removing the external `isEditing` prop dependency. 
* **ProgramExpensesTable:** Updated checkbox logic to disable row selection (`disabled={isAnyRowEditing && !isEditedRow}`) for all other records when a specific row is in edit mode.
* **ProgramExpensesToolbar:** Added `.program-select-disabled` SCSS styles and logic to properly visually disable the filter dropdowns when an inline row edit is active.
* **Tests:** Updated test suites in `ProgramExpensesSection.test.tsx`, `ProgramExpensesTable.test.tsx`, and `ReportAnalytics.test.tsx` to align with the new editing state logic and verify the disabled states of inactive rows.

## How to recreate changes

1. Log in to the Admin panel and navigate to the **Reports** section.
2. Open the **'Програмні витрати'** (Program expenses) tab.
3. Click the **Edit** icon for any record in the list.
4. Verify that the Edit and Delete icons are replaced with Accept and Close icons for that record.
5. Verify that all other records have their checkboxes, Edit, and Delete icons disabled.
6. Verify that the filter dropdowns in the toolbar are disabled.
7. Change the USD/UAH amount and click Accept to ensure the record updates correctly.

## CHECK LIST
- [x] New tests was added or existing was modified
- [ ] Changes has severe impact on users
- [ ] Changes has medium impact on users
- [x] Changes has light impact on users

## Summary by CodeRabbit

- **New Features**
  - Program expense sections now open in editing mode by default.
  - Starting a row edit clears bulk selection and hides the bulk-selection bar.
  - Added retry error notifications when updating or adding program expenses fails.

- **Bug Fixes**
  - Bulk deletion is disabled while a row is being edited.
  - Disabled program filters now provide clearer visual feedback and prevent interaction.
  - Improved empty-state and missing-record behavior for program expenses.

### DIFF
--- a/src/pages/admin/reports/components/program-expenses-section/ProgramExpensesSection.tsx
+++ b/src/pages/admin/reports/components/program-expenses-section/ProgramExpensesSection.tsx
@@ -28,16 +28,15 @@ const INITIAL_PROGRAM_EXPENSES_DATA: ProgramExpensesReadOnlyData = {
 const MAX_PROGRAM_EXPENSE_RECORDS = 4;
 
 interface ProgramExpensesSectionProps {
-    isEditing?: boolean;
     isRowEditMode?: boolean;
     onRowEditModeChange?: (isEditing: boolean) => void;
 }
 
 export const ProgramExpensesSection = ({
-    isEditing = false,
     isRowEditMode: propIsRowEditMode,
     onRowEditModeChange,
 }: ProgramExpensesSectionProps) => {
+    const isEditing = true;
     const adminClient = useAdminClient();
     const { addToast } = useToast();
     const [selectedProgramIds, setSelectedProgramIds] = useState<number[]>([]);
@@ -86,13 +85,6 @@ export const ProgramExpensesSection = ({
                 : validSelectedProgramIds;
         });
     }, [data.programs]);
-
-    useEffect(() => {
-        if (!isEditing) {
-            setIsAddProgramExpenseModalOpen(false);
-            setSelectedRecordIds([]);
-        }
-    }, [isEditing]);
 
     const filteredRecords = useMemo(() => {
         if (selectedProgramIds.length === 0) {
--- a/src/pages/admin/reports/components/program-expenses-section/ProgramExpensesSection.tsx
+++ b/src/pages/admin/reports/components/program-expenses-section/ProgramExpensesSection.tsx
@@ -28,15 +28,16 @@ const INITIAL_PROGRAM_EXPENSES_DATA: ProgramExpensesReadOnlyData = {
 const MAX_PROGRAM_EXPENSE_RECORDS = 4;
 
 interface ProgramExpensesSectionProps {
+    isEditing?: boolean;
     isRowEditMode?: boolean;
     onRowEditModeChange?: (isEditing: boolean) => void;
 }
 
 export const ProgramExpensesSection = ({
+    isEditing = true,
     isRowEditMode: propIsRowEditMode,
     onRowEditModeChange,
 }: ProgramExpensesSectionProps) => {
-    const isEditing = true;
     const adminClient = useAdminClient();
     const { addToast } = useToast();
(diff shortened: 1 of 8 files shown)


src/pages/admin/reports/components/program-expenses-section/ProgramExpensesSection.tsx
```diff
@@ -28,16 +28,15 @@ const INITIAL_PROGRAM_EXPENSES_DATA: ProgramExpensesReadOnlyData = {
...
     isRowEditMode?: boolean;
     onRowEditModeChange?: (isEditing: boolean) => void;
 }
 
 export const ProgramExpensesSection = ({
-    isEditing = false,
     isRowEditMode: propIsRowEditMode,
     onRowEditModeChange,
 }: ProgramExpensesSectionProps) => {
+    const isEditing = true;
```
mehalyna: Since isEditing is now an unconditional constant, every downstream use is dead conditional logic that will mislead future readers into thinking a read-only mode still exists:

- ProgramExpensesSection.tsx:290-294: isAddProgramExpenseDisabled || !isEditing → || false; isEditing ? handleRecordSave : undefined → always handleRecordSave; same for onAddProgramExpense and onDeleteRecord.
- The isEditing prop is still threaded into ProgramExpensesToolbar and ProgramExpensesTable, which keep their own isEditing?: boolean branches (ProgramExpensesTable.tsx colgroup/thead/checkbox column, ProgramExpensesToolbar.tsx:93).

If read-only mode is genuinely gone, delete the ternaries and pass constant values (or drop the isEditing prop from the children too). If it might come back, keep the prop rather than hardcoding. Hardcoding a true const that shadows a just-removed prop is the worst of both — reviewers can't tell whether the read-only path is intentionally dead or accidentally broken.

src/pages/admin/reports/components/program-expenses-section/components/program-expenses-table/ProgramExpensesTable.tsx
```diff
@@ -371,7 +372,7 @@ export const ProgramExpensesTable = ({
                                             <td className={cn(styles.td, styles['checkbox-td'])}>
                                                 <input
                                                     type="checkbox"
-                                                    disabled={isAnyRowEditing}
+                                                    disabled={isAnyRowEditing && !isEditedRow}
```
mehalyna: Previously all row checkboxes were disabled while any row was being edited; now the row being edited stays selectable. Since edit-start clears the selection and the "delete selected" button is disabled (line 306), a user can tick that checkbox mid-edit, which makes the selection bar reappear with a counter and a disabled delete button. That's an odd transient state. If the goal was purely visual (don't grey out the active row), consider styling it as read-only rather than making it interactive. Please confirm this is intended — the commit message only mentions disabling the delete button, not enabling this checkbox.

mehalyna (commented):
Inconsistent mock cleanup in the two new tests: the "update fails" test relies on the new afterEach(jest.restoreAllMocks()) (ProgramExpensesSection.test.tsx:282) while the "add fails" test calls spyPost.mockRestore() explicitly. Both work; pick one style. Also note afterEach restoreAllMocks combined with beforeEach clearAllMocks is fine here, but the redundancy is worth a comment or consolidation.

mehalyna (approved):

PR #494, opened 2026-07-13
Feature/issue 2487 [Reports] Upload PDF file in EN

fhyhtyh:
## Github ticket

* [Github ticket](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2487)

## Description

This PR resolves multiple UX and logic issues in the Admin Panel's "Звітність" (Reports) page within the PDF Files section. Previously, newly uploaded PDF files were appearing at the bottom of the list, retaining their `.pdf` extension in the display name, and causing the entire table to flicker into a full-page loader upon every upload, delete, or rename action. These changes ensure the UI seamlessly update "on the fly" while displaying new files precisely at the top of the list.

## Summary of change

*   **Optimistic UI Updates (`PdfFilesSection.tsx`)**: Removed the redundant `uploadedFiles` array state and simplified optimistic updates to directly mutate `fetchedFiles` (`[newFile, ...prev]`), immediately displaying new uploads at the top. Added corresponding optimistic `totalCount` updates to maintain accurate background pagination alignment for subsequent API calls.
*   **Resolved Table Flicker (`PdfFilesSection.tsx`)**: Introduced the `isInitialFilesLoading` state. The full-page `<InlineLoader />` is now strictly limited to the initial data load or when switching language tabs. Uploads, renames, and deletions now trigger silent background refetches, completely eliminating UI unmounting and flickering.
*   **Test Suite Stabilization (`PdfFilesSection.test.tsx`)**: Updated the `useDataFetch` test mocks to properly implement stateful behavior (using internal `React.useState`) and included the missing `setData` mock. This ensures that optimistic updates trigger re-renders properly and resolves false-negative test failures.

## How to recreate changes

1. Log into the Admin panel and navigate to the **"Звітність"** (Reports) page.
2. Navigate to the **"PDF Файли"** (PDF Files) tab.
3. Select the **EN** language tab.
4. Upload a new PDF file (via drag-and-drop or file dialog).
5. **Verify:** The new file appears *immediately* at the very top of the list without the page or table flickering into a loading spinner.
6. Click to rename or delete an existing file and **verify** that the action completes seamlessly without triggering a full-screen loading state.

## CHECK LIST

## Summary by CodeRabbit

* **Bug Fixes**
  * PDF file lists now reflect uploads, renames, deletions, and reorder actions immediately and consistently.
  * Language switching now fully resets totals and clears the displayed PDF file list.
  * Improved initial loading behavior to better distinguish first-time loading from later refreshes.
* **Tests**
  * Updated the mock data-fetch setup so state changes made by tests propagate through the UI when verifying optimistic updates.

### DIFF
--- a/src/pages/admin/reports/components/pdf-files-section/PdfFilesSection.test.tsx
+++ b/src/pages/admin/reports/components/pdf-files-section/PdfFilesSection.test.tsx
@@ -33,7 +33,16 @@ jest.mock('./components/pdf-section-content-block/PdfSectionContentBlock', () =>
 let capturedOnReorderFiles: ((reordered: any[]) => void) | undefined;
 
 jest.mock('./components/pdf-files-table/PdfFilesTable', () => ({
-    PdfFilesTable: ({ files, onDeleteFile, onViewFile, onRenameFile, onReorderFiles, isDeleting, isRenaming }: any) => {
+    PdfFilesTable: ({
+        files,
+        onDeleteFile,
+        onViewFile,
+        onRenameFile,
+        onReorderFiles,
+        isDeleting,
+        isRenaming,
+        onLoadMore,
+    }: any) => {
         capturedOnReorderFiles = onReorderFiles;
         return (
             <div data-testid="files-table">
@@ -52,13 +61,20 @@ jest.mock('./components/pdf-files-table/PdfFilesTable', () => ({
                 <button onClick={() => onReorderFiles && onReorderFiles(files)} data-testid="reorder-btn">
                     Reorder
                 </button>
+                <button onClick={() => onLoadMore && onLoadMore()} data-testid="load-more-btn">
+                    Load More
+                </button>
             </div>
         );
     },
 }));
 
 jest.mock('./components/language-switcher-buttons/LanguageSwitcherButtons', () => ({
-    LanguageSwitcherButtons: () => <div data-testid="lang-switcher">LanguageSwitcher</div>,
+    LanguageSwitcherButtons: ({ onLanguageChange }: any) => (
+        <button data-testid="lang-switcher" onClick={() => onLanguageChange('en')}>
+            LanguageSwitcher
+        </button>
+    ),
 }));
 
 jest.mock('./components/pdf-dropzone/PdfDropzone', () => ({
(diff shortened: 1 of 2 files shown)


src/pages/admin/reports/components/pdf-files-section/PdfFilesSection.tsx
```diff
@@ -278,18 +270,16 @@ export const PdfFilesSection = () => {
...
-    if (isSectionLoading || isFilesLoading || !activeLanguageId) {
+    const isInitialFilesLoading = isFilesLoading && (!fetchedFiles || fetchedFiles.length === 0) && totalCount === 0;
+
+    if (isSectionLoading || isInitialFilesLoading || !activeLanguageId) {
         return (
             <div className={styles.loader}>
                 <InlineLoader size={3} />
             </div>
         );
     }
```
mehalyna: The new gate is:
```
const isInitialFilesLoading = isFilesLoading && (!fetchedFiles || fetchedFiles.length === 0) && totalCount === 0;
```
The intent — avoid the full-screen loader during background refetches when a list already exists — is good and works for delete/rename/reorder on non-empty lists. But handleDeleteFile optimistically empties the list and resets the count before the follow-up refetch:

```
setFetchedFiles((prev) => (prev ?? []).filter((f) => f.id !== fileId));
setTotalCount((prev) => (prev > 0 ? prev - 1 : 0));
...
await refetchFiles();   // sets isFilesLoading = true
```

When you delete the last remaining file, fetchedFiles.length === 0 and totalCount === 0, so during that refetch isInitialFilesLoading becomes true and the entire section (content block, language switcher, dropzone, table) is replaced by InlineLoader, then snaps back. Same momentary flash on language switch (which also clears both). Consider keying "initial" loading off a dedicated first-load flag/ref rather than inferring it from an empty list + zero count.

src/pages/admin/reports/components/pdf-files-section/PdfFilesSection.tsx
```diff
@@ -161,10 +154,11 @@ export const PdfFilesSection = () => {
     const handleUploaded = useCallback(
         (newFile: PdfReportDto) => {
             loadMoreAbortControllerRef.current?.abort();
-            setUploadedFiles((prev) => [...prev, newFile]);
+            setFetchedFiles((prev) => [newFile, ...(prev ?? [])]);
+            setTotalCount((prev) => prev + 1);
             refetchFiles();
         },
-        [refetchFiles],
+        [refetchFiles, setFetchedFiles],
```
mehalyna, line 173: The optimistic prepend + totalCount + 1 is only reconciled when refetchFiles() succeeds and replaces the list with server data. useDataFetch.fetchData swallows errors internally (sets error, doesn't rethrow), and the call is fire-and-forget. If the refetch fails (network blip), the optimistic entry stays and totalCount remains inflated by 1, which then feeds the handleLoadMore currentLength >= totalCount gate. Minor, but the old code had the same fire-and-forget pattern with a separate array; worth a .catch or awaiting to keep count/list consistent.

src/pages/admin/reports/components/pdf-files-section/PdfFilesSection.test.tsx
```diff
@@ -123,19 +123,22 @@ describe('PdfFilesSection', () => {
...
+            const [data, setData] = React.useState(options.filesData ?? mockFilesResponse.items);
             return {
-                data: options.filesData ?? mockFilesResponse.items,
+                data,
                 isLoading: false,
                 refetch: mockRefetch,
-                setData: options.setFilesData ?? jest.fn(),
+                setData: options.setFilesData ?? setData,
             };
         });
```
mehalyna, line 167:
```
(useDataFetch as jest.Mock).mockImplementation(({ initialData }) => {
    const React = require('react');
    ...
    const [data, setData] = React.useState(...);
```
This makes the mock stateful so optimistic updates reflect in data (that's why the "add uploaded file" test now asserts count 0→1). It works today, but it relies on useDataFetch being invoked exactly twice, in stable order, every render. If the component ever adds/removes/conditionally-calls a useDataFetch, hook order breaks and tests fail confusingly rather than the mock adapting. It's a Rules-of-Hooks landmine embedded in a mock. A more robust approach: return a controllable stateful stub object per instance (e.g. keyed by initialData) without invoking React hooks inside the mock.

mehalyna (commented):
No crashing/data-loss bugs. Finding #1 (last-file delete loader flash) is the one I'd fix before merge; #2 and #3 are worth a quick follow-up.

mehalyna (approved):

PR #498, opened 2026-07-16
Fix/issue-3096 [Reports, Delete Category, Admin Panel] "Програмні" categories are displayed in the "Категорія" dropdown and can be deleted

NikitaPankin:
## Github ticker

* [#3096](https://github.com/ita-social-projects/VictoryCenter-Back/issues/3096)

## Description

This PR resolves an issue where categories whose name contains the reserved "Програмні" label (used by the system aggregate row on the "Доходи та витрати" sub-tab) were shown in the "Категорія" dropdown of both the "Видалити категорію" and "Редагувати категорію" modals — and could be selected for deletion or renamed, even though such categories must be protected from destructive changes. [Related backend PR](https://github.com/ita-social-projects/VictoryCenter-Back/pull/3108)

## How it looks

https://github.com/user-attachments/assets/79e11c49-ca81-4dd9-94f8-33d6fa06ad38

## Summary of change

- **New helper:** added `isReservedFundsExpendituresCategoryName` (`src/utils/functions/is-reserved-funds-expenditures-category-name/`) that flags an expense category as reserved when its name contains the `PROGRAM_EXPENSES_TEXT.TABLE.TYPE_LABEL` ("Програмні") label.
- **`FundsExpendituresSection.tsx`:** introduced `manageableCategories`, a filtered category list (excludes reserved categories) now passed to both `DeleteFundsExpendituresCategoryModal` and `EditFundsExpendituresCategoryModal` instead of the raw category list.
- **`handleDeleteCategory` / `handleEditCategory`:** added a guard that rejects the operation (with the existing error toast) if the target category is reserved, or — for edit — if the requested new name would make it reserved. This protects the flow even if the dropdown filter is bypassed (e.g. stale state).
- **Tests:** added unit tests for the new helper and for the Delete/Edit category dropdown filtering in `FundsExpendituresSection.test.tsx`.

## How to recreate changes

1. Log in as an Administrator.
2. Navigate to the 'Звітність' page.
3. Open the 'Звіт та аналітика' tab.
4. Navigate to the 'Доходи та витрати' sub-tab.
5. Click the burger menu icon next to any category in the table.
6. Select the 'Видалити' (or 'Редагувати') option from the opened menu.
7. Verify: open the 'Категорія' dropdown — no category whose name contains "Програмні" is listed, so it cannot be selected for deletion or editing.

## CHECK LIST

## Summary by CodeRabbit

* **New Features**
  * Added protection for reserved expense categories when creating, editing, or deleting categories.
  * Reserved categories are excluded from edit and delete dropdowns.
  * Added validation feedback for attempts to use reserved category names.
* **Bug Fixes**
  * Improved category management to prevent unintended changes to protected categories.
* **Tests**
  * Expanded coverage for reserved-name validation, category creation, and category filtering.

### DIFF
--- a/src/utils/functions/is-reserved-funds-expenditures-category-name/is-reserved-funds-expenditures-category-name.ts
+++ b/src/utils/functions/is-reserved-funds-expenditures-category-name/is-reserved-funds-expenditures-category-name.ts
@@ -1,0 +1,9 @@
+import { PROGRAM_EXPENSES_TEXT } from '@/const/admin/reports';
+import { FundsExpendituresTransactionType } from '@/types/admin/reports';
+
+export const isReservedFundsExpendituresCategoryName = (
+    name: string,
+    type: FundsExpendituresTransactionType,
+): boolean =>
+    type === 'expense' &&
+    name.trim().toLowerCase().includes(PROGRAM_EXPENSES_TEXT.TABLE.TYPE_LABEL.trim().toLowerCase());
--- a/src/utils/functions/is-reserved-funds-expenditures-category-name/is-reserved-funds-expenditures-category-name.ts
+++ b/src/utils/functions/is-reserved-funds-expenditures-category-name/is-reserved-funds-expenditures-category-name.ts
@@ -6,4 +6,4 @@ export const isReservedFundsExpendituresCategoryName = (
     type: FundsExpendituresTransactionType,
 ): boolean =>
     type === 'expense' &&
-    name.trim().toLowerCase().includes(PROGRAM_EXPENSES_TEXT.TABLE.TYPE_LABEL.trim().toLowerCase());
+    name.trim().toLowerCase().startsWith(PROGRAM_EXPENSES_TEXT.TABLE.TYPE_LABEL.trim().toLowerCase());
--- a/src/const/admin/reports.ts
+++ b/src/const/admin/reports.ts
@@ -261,6 +261,7 @@ export const FUNDS_EXPENDITURES_TEXT = {
             SUBMIT_BUTTON: 'Зберегти',
             ERROR: {
                 NAME_DUPLICATE: 'Категорія з такою назвою вже існує',
+                NAME_RESERVED: 'Ця назва зарезервована системою',
             },
         },
         TRANSLATE_CATEGORY: {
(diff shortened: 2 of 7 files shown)


src/utils/functions/is-reserved-funds-expenditures-category-name/is-reserved-funds-expenditures-category-name.ts
```diff
@@ -0,0 +1,9 @@
+import { PROGRAM_EXPENSES_TEXT } from '@/const/admin/reports';
+import { FundsExpendituresTransactionType } from '@/types/admin/reports';
+
+export const isReservedFundsExpendituresCategoryName = (
+    name: string,
+    type: FundsExpendituresTransactionType,
+): boolean =>
+    type === 'expense' &&
+    name.trim().toLowerCase().includes(PROGRAM_EXPENSES_TEXT.TABLE.TYPE_LABEL.trim().toLowerCase());
```
mehalyna: TYPE_LABEL is 'Програмні', so this treats any expense category whose name contains the substring програмні as reserved. That flags names that aren't the reserved category:

"Непрограмні витрати" (non-program expenses) → "непрограмні".includes("програмні") is true, so it would be silently hidden from the edit/delete dropdowns and blocked from mutation.
Any user-created expense category like "Програмні витрати на IT" gets the same treatment.
The tests ('Програмні тест 2', 'програмні тест') show substring matching is intentional to catch backend-generated variants, so this may be acceptable given the domain — but as written it will also swallow legitimate categories a user creates. If the reserved category is truly a single fixed record, an anchored match (e.g. startsWith, or exact-equality against the known variants) would be safer. Worth confirming with the business rule; flagging as medium-confidence since the substring behavior is asserted by tests.

mehalyna (commented):
FundsExpendituresSection.tsx:
handleEditCategory and handleDeleteCategory both reject reserved names, but handleCreateCategory does not. A user can create an expense category named e.g. "Програмні щось"; it will then be filtered out of manageableCategories and become permanently un-editable and un-deletable through the UI. Depending on whether the backend rejects such names, this could strand a category. Not a regression from this PR, but the guard is asymmetric - consider blocking creation of reserved-looking names too, or relying on the substring narrowing from comment in is-reserved-funds-expenditures-category-name.ts.

mehalyna (approved):

PR #504, opened 2026-07-20
Fix/issue-3112 [Reports] [UI]"Додати витрату" button visual design mismatches Figma mock-up

NikitaPankin:
## JIRA or Github ticker

* [#3112](https://github.com/ita-social-projects/VictoryCenter-Back/issues/3112)

## Description

The `Сума (USD)` field in the `Додати витрату` modal (and the `Додати витрату по програмі` modal, which shares the same logic) in the Reports section showed an inline warning `Сума в USD не відповідає сумі в UAH` when the UAH and USD amounts did not match the exchange rate, but the submit button remained enabled. The button state was derived only from field validation errors and never checked the mismatch warning, so a record with inconsistent amounts could still be submitted.

## How it looks

https://github.com/user-attachments/assets/d04328dd-9e3a-48e6-bf81-71fd8c07acad

## Summary of change

- `useFundsExpendituresRecordForm.ts`: `isSubmitDisabled` now also returns `true` when `usdMismatchMessage` is set, so the submit button is disabled while the UAH and USD amounts do not match the exchange rate.
- `useProgramExpenseRecordForm.ts`: `isSubmitEnabled` now also requires that `usdMismatchMessage` is not set, applying the same fix to the `Додати витрату по програмі` modal.
- `useFundsExpendituresRecordForm.test.ts` and `useProgramExpenseRecordForm.test.ts`: added tests asserting that `isSubmitDisabled` becomes `true` once a UAH/USD mismatch is detected on blur.
- `AddProgramExpenseRecordModal.test.tsx`: updated an existing test that used inconsistent UAH/USD test values unrelated to this scenario, so it no longer triggers the mismatch guard by accident.

## How to recreate changes

1. Log into the Admin panel and navigate to the `Звітність` page.
2. Click `Додати витрату`.
3. Enter `450` into `Сума (UAH)` (the system auto-calculates `Сума (USD)` to `10` based on a rate of `45.00`).
4. Manually change the value in `Сума (USD)` to an inconsistent amount, for example `15`, and remove focus from the field.
5. Verify the message `Сума в USD не відповідає сумі в UAH` appears and the `Додати витрату` submit button becomes disabled.

## CHECK LIST

## Summary by CodeRabbit

* **Bug Fixes**
  * Blocked saving/submission for both expenditure and program expense forms when UAH/USD amounts don’t match the configured exchange rate.
  * Ensured the mismatch warning is correctly shown and the submit button is disabled when a mismatch is detected (including during edit flows with previously stored “legacy” values).

* **Tests**
  * Added and expanded unit test coverage for USD/UАH exchange-rate mismatch scenarios, including cases where confirmation is attempted without a prior USD blur.
  * Updated modal coverage related to submission-pending behavior.

### DIFF
--- a/src/hooks/admin/use-funds-expenditures-record-form/useFundsExpendituresRecordForm.ts
+++ b/src/hooks/admin/use-funds-expenditures-record-form/useFundsExpendituresRecordForm.ts
@@ -199,7 +199,8 @@ export const useFundsExpendituresRecordForm = ({
         Boolean(amountUsdValidationError) ||
         Boolean(categoryValidationError) ||
         Boolean(formState.errors.amountUah) ||
-        Boolean(formState.errors.amountUsd);
+        Boolean(formState.errors.amountUsd) ||
+        Boolean(usdMismatchMessage);
 
     const handleOpenAddConfirmation = useCallback(() => {
         setIsAddConfirmationOpen(true);
--- a/src/hooks/admin/use-funds-expenditures-record-form/useFundsExpendituresRecordForm.ts
+++ b/src/hooks/admin/use-funds-expenditures-record-form/useFundsExpendituresRecordForm.ts
@@ -11,7 +11,7 @@ import {
     validateFundsExpendituresCategory,
     validateFundsExpendituresReportingYear,
 } from '@/validation/admin/reports-schema/funds-expenditures-record-schema/funds-expenditures-record-schema';
-import { useAmountBlur } from '@/hooks/admin/use-amount-blur/useAmountBlur';
+import { getUsdMismatchMessage, useAmountBlur } from '@/hooks/admin/use-amount-blur/useAmountBlur';
 
 interface FundsExpendituresRecordFormState {
     reportingYear: string | undefined;
@@ -157,6 +157,12 @@ export const useFundsExpendituresRecordForm = ({
             return;
         }
 
+        const mismatchMessage = getUsdMismatchMessage(normalizedAmountUah, normalizedAmountUsd, exchangeRate);
+        if (mismatchMessage) {
+            setUsdMismatchMessage(mismatchMessage);
+            return;
+        }
+
         if (!formState.categoryId) {
             return;
         }
@@ -178,7 +184,7 @@ export const useFundsExpendituresRecordForm = ({
         } finally {
             setIsSubmitting(false);
         }
-    }, [formState, getCategoryError, onSubmit, transactionType, setUsdMismatchMessage]);
+    }, [formState, getCategoryError, onSubmit, transactionType, exchangeRate, setUsdMismatchMessage]);
 
     const isDirty =
         Boolean(formState.reportingYear) ||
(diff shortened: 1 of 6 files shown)


src/hooks/admin/use-program-expense-record-form/useProgramExpenseRecordForm.ts
```diff
@@ -223,6 +223,7 @@ export const useProgramExpenseRecordForm = ({
         !getProgramError(formState.programId, 'blur') &&
         !validateProgramExpenseAmount(formState.amountUah, 'save') &&
         !validateProgramExpenseAmount(formState.amountUsd, 'save') &&
+        !usdMismatchMessage &&
```
mehalyna: useProgramExpenseRecordForm supports edit mode (recordToEdit), pre-filling amountUah/amountUsd from a stored record. Those historical amounts were computed against the exchange rate at the time. When the current exchangeRate prop differs, any blur of the pre-filled USD field sets usdMismatchMessage, and the new !usdMismatchMessage term forces isSubmitDisabled = true. An admin editing an unrelated field (e.g. reporting year) can then be blocked from saving a record that was previously legitimate. The funds hook has no edit mode, so this is specific to the program-expense hook.

src/hooks/admin/use-funds-expenditures-record-form/useFundsExpendituresRecordForm.ts
```diff
@@ -199,7 +199,8 @@ export const useFundsExpendituresRecordForm = ({
         Boolean(amountUsdValidationError) ||
         Boolean(categoryValidationError) ||
         Boolean(formState.errors.amountUah) ||
-        Boolean(formState.errors.amountUsd);
+        Boolean(formState.errors.amountUsd) ||
+        Boolean(usdMismatchMessage);
```
mehalyna, line 211: The "block submit on mismatch" invariant now lives in three places: useAmountBlur sets the message, and each of the two form hooks separately folds it into its own differently-shaped isSubmitDisabled expression. The hook that owns the mismatch state (useAmountBlur) doesn't own the submit-blocking decision, so a future third consumer will silently omit the guard. Consider having useAmountBlur expose a derived isMismatch/validate helper (recomputing from current amounts + rate) that both the submit gate and the submit handler call — which would also close finding mentioned in comment to PR

mehalyna (commented):
Mismatch is not re-checked at submit time - useProgramExpenseRecordForm.ts:240, useFundsExpendituresRecordForm.ts:138
handleUsdChange clears usdMismatchMessage on every keystroke, and it's only re-set on USD blur. If a submit path fires without a USD blur in between (programmatic value change → confirm, keyboard flows, or the reset-and-resubmit case), usdMismatchMessage is undefined, the submit gate passes, and onSubmit/submitRecord sends amounts that violate the rate. Neither handleSubmit nor submitRecord independently calls isUsdAmountMismatch, so the guard is UI-timing-dependent rather than enforced at the point of submission. Notably the updated modal test at AddProgramExpenseRecordModal.test.tsx:333 does include a blur, so it passes — but that's the fragile path, not proof the gap is closed.

src/hooks/admin/use-program-expense-record-form/useProgramExpenseRecordForm.ts
NikitaPankin: Added an `areAmountsUnchangedFromRecord` check, mismatch is now only enforced when the admin actually edits an amount field, not on unrelated field edits.

src/hooks/admin/use-funds-expenditures-record-form/useFundsExpendituresRecordForm.ts
NikitaPankin: `useAmountBlur` now exports `getUsdMismatchMessage`, used by both hooks' submit gate and submit handler as the single source of truth.

mehalyna (approved):

PR #561, opened 2026-08-18
Fix/issue-3233-3234 [Admin panel][Reports] Zeros in front of whole numbers are not removed in the 'Сума (UAH)' and 'Сума (USD)'fields

yaroslavkucher:
## JIRA or Github ticker

* [3233](https://github.com/ita-social-projects/VictoryCenter-Back/issues/3233)
* [3234](https://github.com/ita-social-projects/VictoryCenter-Back/issues/3234)

## Description

This PR fixes a bug in the funds expenditures report form where leading zeros were not being removed from the amount input field. The main goal is to ensure that numeric string inputs with leading zeros (e.g., `00000001`) are automatically and correctly normalized to represent the true number (e.g., `1`) upon input formatting.

## Summary of change

* Added a `stripLeadingZeros` helper function using the regex `/^(-?)0+(?=\d)/` inside `funds-expenditures-record-schema.ts`.
* Updated `normalizeFundsExpendituresAmountInput` to apply the stripping logic to the integer part of the amount string before returning the normalized value.
* Handled string formatting for values both with and without decimal comma separators.

## How to recreate changes

1. Log in as an admin.
2. Open the Reports admin page and navigate to the **'Звіт та аналітика'** -> **'Доходи та витрати'** tab.
3. Open the **'Додати надходження'** modal.
4. Enter a value with leading zeros in the amount field (e.g., `00000001`).
5. Focus out of the field (trigger the blur event).
6. Observe the field: all zeros in front of the `1` are successfully removed. No validation error messages are shown.

## CHECK LIST


d10max (approved):

cirillad:
You need to pull the latest changes from the release branch into this fix branch, because Sonar might be checking the difference between the target branch and the merging branch.

cirillad (approved):

src/validation/admin/reports-schema/funds-expenditures-record-schema/funds-expenditures-record-schema.ts
```diff
@@ -20,9 +20,10 @@ export const normalizeFundsExpendituresAmountInput = (value: string | undefined
     }
     const withNormalizedSpaces = value.replaceAll(/\s+/g, ' ').trimStart();
     let withCommaSeparator = withNormalizedSpaces.replaceAll('.', ',');
-
     withCommaSeparator = withCommaSeparator.replace(/^-\s+/, '-');
 
+    const stripLeadingZeros = (str: string) => str.replace(/^(-?)0+(?=\d)/, '$1');
```
mehalyna: The regex pattern ^(-?)0+(?=\d) does not handle the case where the string is only zeros (e.g., "00000"). Apply the regex and then check if the result is empty, returning "0" as a fallback. Consider adding: `const result = str.replace(/^(-?)0+(?=\d)/, '$1'); return result === '' || result === '-' ? '0' : result;`

src/validation/admin/reports-schema/funds-expenditures-record-schema/funds-expenditures-record-schema.test.ts
```diff
@@ -25,6 +25,25 @@ describe('FUNDS_EXPENDITURES_RECORD_VALIDATION_FUNCTIONS', () => {
             expect(normalizeFundsExpendituresAmountInput('1200,5678')).toBe('1200,56');
         });
 
+        it('should remove leading zeros from the integer part', () => {
+            expect(normalizeFundsExpendituresAmountInput('00000001')).toBe('1');
+            expect(normalizeFundsExpendituresAmountInput('0050')).toBe('50');
+        });
+
+        it('should leave a single zero if the integer part is only zeros', () => {
+            expect(normalizeFundsExpendituresAmountInput('00000')).toBe('0');
```
mehalyna, line 34: The test expects '00000' to normalize to '0', but based on the regex implementation, it will return an empty string. Verify that the implementation correctly handles inputs that are all zeros and update if necessary.

src/validation/admin/reports-schema/funds-expenditures-record-schema/funds-expenditures-record-schema.ts
mehalyna: The helper function is defined inside the export function but is redefined on every call to normalizeFundsExpendituresAmountInput. Move this function outside as a private constant or extract it to improve performance and maintainability.

src/validation/admin/reports-schema/funds-expenditures-record-schema/funds-expenditures-record-schema.test.ts
```diff
@@ -25,6 +25,25 @@ describe('FUNDS_EXPENDITURES_RECORD_VALIDATION_FUNCTIONS', () => {
...
+            expect(normalizeFundsExpendituresAmountInput('0')).toBe('0');
+        });
+
+        it('should remove leading zeros but preserve decimal parts', () => {
+            expect(normalizeFundsExpendituresAmountInput('0005,50')).toBe('5,50');
+            expect(normalizeFundsExpendituresAmountInput('0000,50')).toBe('0,50');
+        });
+
+        it('should handle leading zeros with dot separator correctly', () => {
+            expect(normalizeFundsExpendituresAmountInput('0012.34')).toBe('12,34');
```
mehalyna, line 44: The test expects a dot (.) to be converted to a comma (,), but this is only partially covered in the implementation. Ensure that all decimal separator conversions happen before the leading zeros stripping logic is applied.

src/validation/admin/reports-schema/funds-expenditures-record-schema/funds-expenditures-record-schema.ts
```diff
@@ -32,10 +33,12 @@ export const normalizeFundsExpendituresAmountInput = (value: string | undefined
...
 
     if (firstCommaIndex === -1) {
-        return trimEnd ? withCommaSeparator.trim() : withCommaSeparator;
+        const noZeros = stripLeadingZeros(withCommaSeparator);
+        return trimEnd ? noZeros.trim() : noZeros;
     }
 
-    const integerPart = withCommaSeparator.slice(0, firstCommaIndex);
+    const rawIntegerPart = withCommaSeparator.slice(0, firstCommaIndex);
+    const integerPart = stripLeadingZeros(rawIntegerPart);
```
mehalyna, line 41: When the integer part becomes empty after stripping leading zeros from a value like "0", the code returns an empty string instead of "0". Ensure the stripLeadingZeros function or its usage handles the zero-only case explicitly.

yaroslavkucher:
@mehalyna
Regarding `stripLeadingZeros` position — totally agree, I moved it outside the exported function.

Regarding the regex /^(-?)0+(?=\d)/: the (?=\d) is a positive lookahead that ensures we only match zeros that are followed by at least one other digit. This guarantees that "00000" becomes "0", and a single "0" remains "0". It never returns an empty string for zero-only values. I've also added unit tests in this PR that verify these exact cases ('00000' -> '0', '0' -> '0').

Also, the conversion of . to , happens on line 24 (`withCommaSeparator`), which is executed strictly before `stripLeadingZeros` is called, so decimal separators are fully handled beforehand.

mehalyna (approved):

PR #562, opened 2026-08-18
Fix/issue 3001 (Pdf reports do not update on the fly when admin modifies them through the admin panel)

g1temn:
Environment:
Windows 10 
Chrome 151.0.7922.174

Issue ticket link
Github ticket: https://github.com/ita-social-projects/VictoryCenter-Back/issues/3001

Summary of issue:
The newly uploaded PDF file does not appear on the Public 'Звітність' page automatically. It only becomes visible after a manual page reload which is not a correct behaviour according to the user story. #1326

Summary of change:
SignalR library was implemented in a form of a custom hook that lets pdf reports component to listen to incoming messages about admin updating them.

How to recreate changes
1. Open admin panel => page "Звіти" => tab "Звіт та аналітика"  => subtab "PDF Файли"
2. Open the "Звітність" page on the public side of application
3. Make changes to reports through the admin panel
4. See changes to appear on the public side without a manual reload

How to reproduce changes locally:
1. Switch both your local back and front to the branch fix/issue-3001
2. Run both of them
3. As an admin open your panel => page "Звіти" => tab "Звіт та аналітика"  => subtab "PDF Файли"
4. Simultaniously open the "Звітність" page on the public side of application as a default user
5. Make changes to reports through the admin panel
6. See changes to appear on the public side without a manual reload

CHECK LIST
- [+]  Changes has medium impact on users
- [+]  PR meets all conventions

### DIFF
--- a/src/hooks/public/SignalR/useSignalR.ts
+++ b/src/hooks/public/SignalR/useSignalR.ts
@@ -1,0 +1,25 @@
+import { useState, useEffect } from 'react';
+import { HubConnection, HubConnectionBuilder } from '@microsoft/signalr';
+
+export const useSignalR = (url: string): HubConnection | null => {
+    const [connection, setConnection] = useState<HubConnection | null>(null);
+
+    useEffect(() => {
+        const newConnection = new HubConnectionBuilder().withUrl(url).withAutomaticReconnect().build();
+
+        newConnection
+            .start()
+            .then(() => {
+                setConnection(newConnection);
+            })
+            .catch((err: Error) => {
+                console.error('SignalR Connection Error: ', err);
+            });
+
+        return () => {
+            newConnection.stop();
+        };
+    }, [url]);
+
+    return connection;
+};
--- a/src/hooks/public/SignalR/useSignalR.ts
+++ b/src/hooks/public/SignalR/useSignalR.ts
@@ -1,22 +1,39 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import { HubConnection, HubConnectionBuilder } from '@microsoft/signalr';
 
 export const useSignalR = (url: string): HubConnection | null => {
     const [connection, setConnection] = useState<HubConnection | null>(null);
+    const isMounted = useRef(true);
 
     useEffect(() => {
+        isMounted.current = true;
+
         const newConnection = new HubConnectionBuilder().withUrl(url).withAutomaticReconnect().build();
 
-        newConnection
-            .start()
-            .then(() => {
-                setConnection(newConnection);
-            })
-            .catch((err: Error) => {
-                console.error('SignalR Connection Error: ', err);
-            });
+        let retryCount = 0;
+        const maxRetries = 5;
+
+        const startConnection = async () => {
+            try {
+                await newConnection.start();
+                if (isMounted.current) {
+                    setConnection(newConnection);
+                }
+            } catch (err) {
+                console.error(`SignalR Connection Error (Attempt ${retryCount + 1}): `, err);
+
+                if (retryCount < maxRetries && isMounted.current) {
+                    const timeout = Math.pow(2, retryCount) * 1000;
+                    retryCount++;
+                    setTimeout(startConnection, timeout);
+                }
+            }
+        };
+
+        startConnection();
 
         return () => {
+            isMounted.current = false;
             newConnection.stop();
         };
     }, [url]);
--- a/src/hooks/public/SignalR/useSignalR.ts
+++ b/src/hooks/public/SignalR/useSignalR.ts
@@ -20,8 +20,6 @@ export const useSignalR = (url: string): HubConnection | null => {
                     setConnection(newConnection);
                 }
             } catch (err) {
-                console.error(`SignalR Connection Error (Attempt ${retryCount + 1}): `, err);
-
                 if (retryCount < maxRetries && isMounted.current) {
                     const timeout = Math.pow(2, retryCount) * 1000;
                     retryCount++;
--- a/src/pages/public/reports-page/components/reports-section/ReportsSection.tsx
+++ b/src/pages/public/reports-page/components/reports-section/ReportsSection.tsx
@@ -55,10 +55,10 @@ export const ReportsSection = () => {
             }
         };
 
-        signalRConnection.on('PdfReportCreated', handlePdfCreated);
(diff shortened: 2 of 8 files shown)


src/hooks/public/SignalR/useSignalR.ts
```diff
@@ -0,0 +1,25 @@
+import { useState, useEffect } from 'react';
+import { HubConnection, HubConnectionBuilder } from '@microsoft/signalr';
+
+export const useSignalR = (url: string): HubConnection | null => {
+    const [connection, setConnection] = useState<HubConnection | null>(null);
+
+    useEffect(() => {
+        const newConnection = new HubConnectionBuilder().withUrl(url).withAutomaticReconnect().build();
```
mehalyna, line 11: The HubConnection is created and started without any error handling for network failures during initialization. If the connection fails to establish, the component will silently fall back to null and never retry, leaving the real-time updates completely non-functional. Add a retry mechanism with exponential backoff to handle transient network failures.

src/pages/public/reports-page/components/reports-section/ReportsSection.tsx
```diff
@@ -16,6 +17,8 @@ export const ReportsSection = () => {
     const [isExpanded, setIsExpanded] = useState(false);
     const [reports, setReports] = useState<PdfReportDto[]>([]);
 
+    const signalRConnection = useSignalR(process.env.SIGNALR_PDF_REPORTS_URL || 'https://localhost:5001/hubs/reports');
```
mehalyna: The SignalR URL is hardcoded with a localhost fallback which will fail in production environments. The environment variable SIGNALR_PDF_REPORTS_URL is undefined in .env.development, making real-time updates non-functional in development unless manually configured. Add a non-empty default or clarify the required configuration in environment setup documentation.

src/pages/public/reports-page/components/reports-section/ReportsSection.tsx
```diff
@@ -34,6 +37,31 @@ export const ReportsSection = () => {
...
+        if (!signalRConnection) return;
+
+        const handlePdfCreated = async (updatedLanguageId: number) => {
+            try {
+                const languages = await localizationLanguagesDataFetch();
+                const langCode = currentLanguage?.startsWith('en') ? 'en' : 'uk';
+                const language = languages.find((l) => l.code === langCode);
+
+                if (language && language.id === updatedLanguageId) {
+                    const result = await PdfReportsApi.getAllByLanguageId(language.id, { offset: 0, limit: 1000 });
```
mehalyna: The API call fetches all reports with limit: 1000 which assumes there are never more than 1000 reports and will silently truncate data if this limit is exceeded. Use pagination or dynamic limits based on the actual number of reports.

src/pages/public/reports-page/components/reports-section/ReportsSection.tsx
```diff
@@ -34,6 +37,31 @@ export const ReportsSection = () => {
...
+                console.error('Failed to refetch reports on SignalR event', err);
+            }
+        };
+
+        signalRConnection.on('PdfReportAction', handlePdfCreated);
+
+        return () => {
+            signalRConnection.off('PdfReportAction', handlePdfCreated);
+        };
+    }, [signalRConnection, currentLanguage]);
```
mehalyna: The signalRConnection and currentLanguage dependency array will cause the listener to be re-registered whenever either changes, but signalRConnection updates on every re-render if the hook doesn't stabilize its return value. This creates potential memory leaks and duplicate event listeners. Stabilize the connection reference using useMemo or useCallback to prevent unnecessary re-registrations.

src/pages/public/reports-page/components/reports-section/ReportsSection.tsx
```diff
@@ -34,6 +37,31 @@ export const ReportsSection = () => {
...
+        const handlePdfCreated = async (updatedLanguageId: number) => {
+            try {
+                const languages = await localizationLanguagesDataFetch();
+                const langCode = currentLanguage?.startsWith('en') ? 'en' : 'uk';
+                const language = languages.find((l) => l.code === langCode);
+
+                if (language && language.id === updatedLanguageId) {
+                    const result = await PdfReportsApi.getAllByLanguageId(language.id, { offset: 0, limit: 1000 });
+                    setReports(result.items);
+                }
```
d10max: Since `handlePdfCreated` containts asynchronous API calls, there's a risk of updating an unmounted component. If the user navigates away from this page while these requests are still pending the component will unmount, but the promises will eventually resolve and call `setReports`, leading to a memory leak. Could we introduce `mounted` flag inside the useEffect to check if the component is still mounted before calling state setters, similar to how it's done inside the `useEffect` on lines 22–38.?

src/pages/public/reports-page/components/reports-section/ReportsSection.tsx
d10max: We trigger a full refetch of up to 1000 reports every time a single new report is created. It could lead to heavy database load and performance issues, especially if reports are generated frequently. Would it be possible for the backend to send the newly created report object directly with the SignalR event? If we receive the new report right from the event, we can completely eliminate this API call and just update our local state immutably: `setReports(prev => [newReport, ...prev])`.

g1temn:
all the abovementioned remarks have been taken into account and fixed in the last two commits

summary of change listed in the commit message:
1) SignalR URL is taken from the environment variable now instead of being hardcoded
2) Pagination is now used to fetch pdf reports by portions instead of loading everything at once
3) 'mounted' flag inside of the 'ReportsSection' component was introduced to eliminate the chance of potential memory leaks
4) every possible event that updates the state of pdf reports is now being listened to separately in order to update the 'ReportsSection' component more efficiently
5) error handling and retrying logic of an initially not established connection was introduced in the 'useSignalR.ts' hook

also i added another environment variable (REACT_APP_SIGNALR_PDF_REPORTS_URL) to the README.md so that the prod can use the implemented feature

d10max (commented):
Looks good, will approve after sonar fixes.

Denys1310:
Thank you for your fix! Unfortunately, I couldn't reproduce the changes. Could you please provide screenshots or a video in the PR section "How to reproduce changes"? Additionally, could you provide some information about the environment you used? Thank you!

mehalyna (approved):

Denys1310 (approved):

PR #564, opened 2026-08-18
Fix/issue 3376 [Admin panel][Profile company] Tooltip text is displayed on click instead of hover

Zest005:
## Github ticket

* https://github.com/ita-social-projects/VictoryCenter-Back/issues/3376

## Description

The info tooltip in the admin panel was displayed only after clicking the icon. According to the User Story it must appear on hover and on keyboard focus, without requiring a click.

## How it looks

https://github.com/user-attachments/assets/0c383be2-bcb6-407b-b56c-48cb8e808f74

## Summary of change

`ButtonTooltip.tsx` - the tooltip is now shown by onMouseEnter and onFocus, and hidden by onMouseLeave and onBlur. The click handler was kept for touch devices, but it no longer toggles: it only shows the tooltip. Toggling on click caused the tooltip to disappear while the cursor was still over the icon. useOnClickOutside was kept - it is what closes the tooltip after a tap on touch devices.

`ButtonTooltip.scss` - the icon kept its highlighted background after a click, because a clicked button retains focus and :focus styling stayed applied. Replaced :focus with :focus-visible, so the highlight is cleared once the cursor leaves. :focus-visible was chosen over removing the rule entirely, so the icon still has a visible state when reached with Tab.

`ButtonTooltip.test.tsx` - four tests added for hover and focus. One test was updated because click no longer toggles, and one duplicate test was removed.

## How to recreate changes

1. Open the admin panel -> Company Profile
2. Hover over any info icon next to a field label -> the tooltip appears immediately
3. Move the cursor away -> the tooltip disappears
4. Press Tab until the icon is focused -> the tooltip appears
5. Click the icon and move the cursor away -> the icon returns to its normal state

## CHECK LIST

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tooltip behavior for mouse and keyboard interactions.
  * Tooltips now remain visible after repeated clicks instead of unexpectedly toggling closed.
  * Keyboard focus styling is applied only when focus is visibly present.
  * Tooltips continue to close when focus or pointer interaction ends.

### DIFF
--- a/src/components/admin/button-tooltip/ButtonTooltip.test.tsx
+++ b/src/components/admin/button-tooltip/ButtonTooltip.test.tsx
@@ -1,5 +1,5 @@
 import '@testing-library/jest-dom';
-import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import { render, screen, fireEvent, waitFor, act } from '@testing-library/react';
 import { ButtonTooltip, ButtonTooltipProps } from './ButtonTooltip';
 import { TooltipProps } from '../tooltip/Tooltip';
 
@@ -51,8 +51,14 @@ describe('ButtonTooltip', () => {
 
     const hoverButton = () => fireEvent.mouseEnter(getButton());
     const unhoverButton = () => fireEvent.mouseLeave(getButton());
-    const focusButton = () => fireEvent.focus(getButton());
-    const blurButton = () => fireEvent.blur(getButton());
+    const focusButton = () =>
+        act(() => {
+            getButton().focus();
+        });
+    const blurButton = () =>
+        act(() => {
+            getButton().blur();
+        });
 
     afterEach(() => {
         jest.clearAllMocks();
--- a/src/components/admin/button-tooltip/ButtonTooltip.test.tsx
+++ b/src/components/admin/button-tooltip/ButtonTooltip.test.tsx
@@ -49,6 +49,11 @@ describe('ButtonTooltip', () => {
     const clickTooltip = () => fireEvent.mouseDown(getTooltip()!);
     const clickOutside = () => fireEvent.mouseDown(document.body);
 
+    const hoverButton = () => fireEvent.mouseEnter(getButton());
+    const unhoverButton = () => fireEvent.mouseLeave(getButton());
+    const focusButton = () => fireEvent.focus(getButton());
+    const blurButton = () => fireEvent.blur(getButton());
+
     afterEach(() => {
         jest.clearAllMocks();
     });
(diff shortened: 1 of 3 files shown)


maromexe (approved):
checked locally, hover and keyboard focus both work, and the 
:focus-visible change fixes the highlight staying after a click

src/components/admin/button-tooltip/ButtonTooltip.tsx
```diff
@@ -15,15 +15,19 @@ export const ButtonTooltip = ({ children, position = 'bottom', isRenderInPortal
...
 
     const closeTooltip = useCallback(() => {
         setIsVisible(false);
     }, []);
 
+    const handleClick = (e: React.MouseEvent) => {
+        e.stopPropagation();
+        showTooltip();
+    };
+
```
mehalyna, line 30: The handleClick function calls showTooltip() but is defined after showTooltip in the code, creating a forward reference. Move handleClick before the JSX or inline it to ensure proper function hoisting and readability.

src/components/admin/button-tooltip/ButtonTooltip.test.tsx
```diff
@@ -49,6 +49,11 @@ describe('ButtonTooltip', () => {
     const clickTooltip = () => fireEvent.mouseDown(getTooltip()!);
     const clickOutside = () => fireEvent.mouseDown(document.body);
 
+    const hoverButton = () => fireEvent.mouseEnter(getButton());
+    const unhoverButton = () => fireEvent.mouseLeave(getButton());
+    const focusButton = () => fireEvent.focus(getButton());
```
mehalyna: The helper function focusButton uses fireEvent.focus() which does not simulate real focus behavior consistently across browsers. Use userEvent.tab() or element.focus() directly for more realistic keyboard focus testing.

src/components/admin/button-tooltip/ButtonTooltip.tsx
Zest005: > The handleClick function calls showTooltip() but is defined after showTooltip in the code, creating a forward reference. Move handleClick before the JSX or inline it to ensure proper function hoisting and readability.

showTooltip is declared on line 18 and handleClick on line 26, so handleClick already comes after the function it calls. The return with the JSX starts on line 37, so handleClick is above it as well. Forward reference would be the opposite order. Am I missing something?
handleClick is only called from the click handler (not during render) it shouldn't affect hoisting.
Could you clarify what exactly should be reordered? I can also inline the handler if you prefer.

src/components/admin/button-tooltip/ButtonTooltip.test.tsx
Zest005: > The helper function focusButton uses fireEvent.focus() which does not simulate real focus behavior consistently across browsers. Use userEvent.tab() or element.focus() directly for more realistic keyboard focus testing.

Thanks for correction. Applied and wrapped in act() so the React flushes the state update before the assertion. By the way the same is used in other tests (e.g in `ImageInput.test.tsx` and `InputWithCharacterLimit.test.tsx`). I think that we need a separate task to fix them too.

mehalyna (approved):

PR #567, opened 2026-08-19
Fix/issue-3407 [Admin Panel][Team] Character counter position in the name field

maromexe:
## JIRA or Github ticker

https://github.com/ita-social-projects/VictoryCenter-Back/issues/3407

## Description

The character counter for the 'Назва' field was inside the input 
border, while the 'Опис' counter is correctly positioned below the textarea. 
According to Figma mockup both should be below the field.

## How it looks

image

## Summary of change

`InputWithCharacterLimitGroup` already supports a `showCounterBelow` prop, 
which renders the counter below the field instead of inside it. It just 
wasn't passed in `TeamCategoryModal`, so the default (false) kept the 
counter inside the input.

## How to recreate changes

1. Log in as admin
2. Navigate to the Team page
3. Open the 'Редагувати категорію' modal
4. The counter for 'Назва' is now below the input, aligned with 'Опис'
5. Clear the field - the counter still shows 0/20 below the input

## CHECK LIST

### DIFF
--- a/src/pages/admin/team/components/team-category-modal/TeamCategoryModal.tsx
+++ b/src/pages/admin/team/components/team-category-modal/TeamCategoryModal.tsx
@@ -295,6 +295,7 @@ export const TeamCategoryModal = (props: TeamCategoryModalProps) => {
 
                         <InputWithCharacterLimitGroup
                             isRequired
+                            showCounterBelow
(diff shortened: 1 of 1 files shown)


maromexe:
Closing because it is already fixed in `release/1.0.0`. My branch was based on an 
older state, so this added a duplicate `showCounterBelow` prop

PR #569, opened 2026-08-19
Feature/issue 3316  [Hippotherapy] Create layout of Hippotherapy page

fhyhtyh:
## Github ticket

* [Github ticket](https://github.com/ita-social-projects/VictoryCenter-Back/issues/3316)

## Description

This PR adds the "Іпотерапія" (Hippotherapy) admin page, covering all 11 sections in the same order they render publicly — intro banner, three text-card sections, two quote sections, the Hippovention-center section, two image-gallery sections, the scientific references list, and the ethics section.

## Summary of change

*   **Page & routing (`HippotherapyPageAdmin.tsx`, `routes.ts`, `AdminNavigation.tsx`, `AppRouter.jsx`)**: New `/admin-panel/hippotherapy` route and **"Іпотерапія"** nav link, following the existing admin page pattern.
*   **Content sections (`components/sections/*`)**: 7 section components — `IntroBannerSection`, `TextCardSection` (reused ×3), `HippotherapyQuoteSection` (reused ×2), `HippoventionCenterSection`, `GallerySection` (reused ×2), `ScientificReferencesSection`, `EthicsSection` — covering all 11 public-page content blocks, each with rich-text fields, image upload/crop, and character-limit validation.
*   **Page orchestration (`HippotherapyPageContent.tsx`)**: Fetch/dirty-check/publish flow, with a `404` → blank-form fallback so a first-time page can be created directly from the admin (the `HippotherapyPage` endpoint upserts on save).
*   **API & validation (`hippotherapy-page-api.ts`, `HippotherapyPageSchema.ts`, `types/admin/hippotherapy-page.ts`)**: GET/PUT service against `HippotherapyPage`, image upload/delete resolution, and publish-gating validation (required text + required images per section).
*   **Scientific references list**: Add / expand / delete controls are present as placeholders in this PR.
*   **Test Coverage Expansion (`*.test.tsx`/`*.test.ts`)**: Full test coverage across all new section components, the page container, the API service, and the validation schema, keeping project coverage thresholds intact.

## How to recreate changes

1. Log into the Admin panel and navigate to **"Іпотерапія"**.
2. If the page has no content yet, a blank form appears with an empty-state notice — fill in the required title/description fields for each section and upload the required images.
3. Click **"Опублікувати"** (Publish), confirm the modal, and verify the success toast appears.
5. Reload the page and confirm the content persists and re-populates every section correctly.

## CHECK LIST


mehalyna (approved):
TODO:
hippotherapy-page-api.ts line 13-19
```
const resolveImageValue = async <T extends HippotherapyImageValue>(
    client: AxiosInstance,
    value: T,
    imagesToDelete: number[],
): Promise<T> => {
    if (!value.image && !value.imageId) {
        return value;
    }
    // ...
```
Data loss risk
If a user uploads an image (now in value.image), saves it (image gets imageId), then later deletes it (sets value.image = null but keeps imageId from DB), this code returns early without deleting the image from the server. The image ID is never added to imagesToDelete.
Expected Flow:
User uploads image → image set, imageId = null
Save → API generates imageId, image stored
User deletes image → image = null, but code should still detect imageId exists and delete it
Missing Logic: Check if !value.image && value.imageId (image was deleted but still has ID), then add to imagesToDelete.

Fix Direction: Add condition before early return to handle image deletion transition

maromexe (approved):
Ran this locally and layout looks solid

Since text fields were added here as well, a few things do not match 
the user stories. 2249, 2250, 2251 are on me, so I will fix them 
in my phase rather than block this:
- validation runs on change, so an empty page shows "Поле обов'язкове" 
under every field on load (other admin forms validate on blur)
- "Додатковий опис" renders as required, but the field table in #2249 
marks it as Optional
- several char limits don't match the table e.g. "Додатковий опис" in 
the Hippovention-center section is 50 there, 300 here

stkossman (approved):
Before merge resolve all conflicts

PR #575, opened 2026-08-23
Fix/issue-3446 [Team] [Admin Panel] Scrollbar appears on dropdown list on "Видалити категорію" and "Редагувати категорію" pop ups

momaksymenko:
## Github ticket

* Github ticket [#3446](https://github.com/ita-social-projects/VictoryCenter-Back/issues/3446#issue-5202648849)

## Description

This PR fixes the issue where expanding the category dropdown in "Видалити категорію" and "Редагувати категорію" pop-ups causes an unwanted vertical scrollbar on the modal window frame and crops option elements.

## How it looks

Знімок екрана 2026-08-23 115907
Знімок екрана 2026-08-23 115925

## Summary of change

- Updated TeamCategoryModal.scss to configure absolute positioning and custom scrollbar styles for the category dropdown options
- Updated DeleteTeamCategoryModal.scss to limit dropdown height to 120px with internal scrolling and adjust container spacing to prevent modal scrolling

## How to recreate changes

1. Log in to the Admin Panel.
2. Navigate to the Team page.
3. Open the burger menu and click either 'Редагувати' or 'Видалити' option.
4. Click on the 'Категорія' dropdown list.
5. Expected result: Dropdown list displays fully with an internal custom scrollbar limited to 120px height without triggering a vertical scrollbar on the modal frame itself.

## CHECK LIST

### DIFF
--- a/src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
+++ b/src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
@@ -13,4 +13,35 @@
     display: flex;
     flex-direction: column;
     gap: 0.4375rem;
+    overflow-y: visible;
+
+    .singleselect-options {
+        position: absolute;
+        z-index: 100;
+        max-height: 120px;
+        overflow-y: auto;
+        scrollbar-color: #a4a49b transparent;
+
+        &::-webkit-scrollbar {
+            width: 8px;
+        }
+
+        &::-webkit-scrollbar-track {
+            background: transparent;
+            border-radius: 8px;
+        }
+
+        &::-webkit-scrollbar-thumb {
+            background-color: #a4a49b;
+            border-radius: 8px;
+
+            &:hover {
+                background-color: #8c8c84;
+            }
+        }
+    }
+
+    .input-group:has(.singleselect-options) {
+        margin-bottom: 126px;
+    }
 }
--- a/src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
+++ b/src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
@@ -44,7 +44,7 @@ $dropdown-gap: 6px;
             border-radius: 8px;
 
             &:hover {
-                background-color: colors.$grey-600;
+                background-color: colors.$grey-800;
             }
         }
     }
--- a/src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
+++ b/src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
@@ -1,5 +1,8 @@
 @use '@/assets/sass/variables/colors';
 
+$dropdown-max-height: 120px;
+$dropdown-gap: 6px;
+
 .team-category-modal-error-container {
     font-size: 14px;
     line-height: 20px;
@@ -13,12 +16,23 @@
     display: flex;
     flex-direction: column;
     gap: 0.4375rem;
-    overflow-y: visible;
+    overflow-y: hidden;
+
+    .input-group {
+        position: relative;
+
+        &:has(.singleselect-options) {
+            margin-bottom: calc(#{$dropdown-max-height} + #{$dropdown-gap}); 
+        }
+    }
 
     .singleselect-options {
         position: absolute;
+        top: 100%;
+        left: 0;
+        right: 0;
         z-index: 100;
-        max-height: 120px;
+        max-height: $dropdown-max-height;
         overflow-y: auto;
         scrollbar-color: #a4a49b transparent;
 
@@ -39,9 +53,5 @@
                 background-color: #8c8c84;
             }
         }
-    }
-
-    .input-group:has(.singleselect-options) {
-        margin-bottom: 126px;
     }
 }
--- a/src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
+++ b/src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
@@ -19,22 +19,16 @@ $dropdown-gap: 6px;
     overflow-y: hidden;
 
     .input-group {
-        position: relative;
-
         &:has(.singleselect-options) {
             margin-bottom: calc(#{$dropdown-max-height} + #{$dropdown-gap}); 
         }
     }
 
     .singleselect-options {
-        position: absolute;
-        top: 100%;
-        left: 0;
-        right: 0;
         z-index: 100;
         max-height: $dropdown-max-height;
         overflow-y: auto;
-        scrollbar-color: #a4a49b transparent;
+        scrollbar-color: colors.$grey-700 transparent;
 
         &::-webkit-scrollbar {
             width: 8px;
@@ -46,11 +40,11 @@ $dropdown-gap: 6px;
         }
 
         &::-webkit-scrollbar-thumb {
-            background-color: #a4a49b;
+            background-color: colors.$grey-700;
             border-radius: 8px;
 
             &:hover {
-                background-color: #8c8c84;
+                background-color: colors.$grey-600;
             }
         }
     }
--- a/src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
+++ b/src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
@@ -6,9 +6,19 @@
         flex-direction: column;
         gap: 0.375rem;
 
+        .input-group {
+            position: relative;
+        }
+
         .singleselect-options {
+            position: absolute;
+            top: 100%;
+            left: 0;
+            right: 0;
+            z-index: 100;
             max-height: 120px;
             overflow-y: auto;
+            overflow-x: hidden;
             scrollbar-color: #a4a49b transparent;
 
             &::-webkit-scrollbar {
--- a/src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
+++ b/src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
@@ -6,20 +6,12 @@
         flex-direction: column;
         gap: 0.375rem;
 
-        .input-group {
-            position: relative;
-        }
-
         .singleselect-options {
-            position: absolute;
-            top: 100%;
-            left: 0;
-            right: 0;
             z-index: 100;
             max-height: 120px;
             overflow-y: auto;
             overflow-x: hidden;
-            scrollbar-color: #a4a49b transparent;
+            scrollbar-color: colors.$grey-700 transparent;
 
             &::-webkit-scrollbar {
                 width: 8px;
@@ -31,11 +23,11 @@
             }
 
             &::-webkit-scrollbar-thumb {
-                background-color: #a4a49b;
+                background-color: colors.$grey-700;
                 border-radius: 8px;
 
                 &:hover {
-                    background-color: #8c8c84;
+                    background-color: colors.$grey-600;
                 }
             }
         }
--- a/src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
+++ b/src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
@@ -27,7 +27,7 @@
                 border-radius: 8px;
 
                 &:hover {
-                    background-color: colors.$grey-600;
+                    background-color: colors.$grey-800;
                 }
             }
         }
--- a/src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
+++ b/src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
@@ -5,6 +5,30 @@
         display: flex;
         flex-direction: column;
         gap: 0.375rem;
+
+        .singleselect-options {
+            max-height: 120px;
+            overflow-y: auto;
+            scrollbar-color: #a4a49b transparent;
+
+            &::-webkit-scrollbar {
+                width: 8px;
+            }
+
+            &::-webkit-scrollbar-track {
+                background: transparent;
+                border-radius: 8px;
+            }
+
+            &::-webkit-scrollbar-thumb {
+                background-color: #a4a49b;
+                border-radius: 8px;
+
+                &:hover {
+                    background-color: #8c8c84;
+                }
+            }
+        }
     }
 
     &-modal-error-container {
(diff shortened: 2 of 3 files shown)


src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
```diff
@@ -13,4 +13,35 @@
     display: flex;
     flex-direction: column;
     gap: 0.4375rem;
+    overflow-y: visible;
```
mehalyna: The overflow-y: visible declaration will not prevent the modal scrollbar from appearing. Use overflow-y: hidden on the container instead to prevent scrolling when the dropdown extends beyond the modal bounds.

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
```diff
@@ -13,4 +13,35 @@
...
+            border-radius: 8px;
+
+            &:hover {
+                background-color: #8c8c84;
+            }
+        }
+    }
+
+    .input-group:has(.singleselect-options) {
+        margin-bottom: 126px;
```
mehalyna: Using a fixed 126px margin-bottom on the input-group is a brittle approach that assumes the scrollbar height is always exactly 120px plus 6px gap. Use a more flexible approach with CSS Grid or adjust the container's max-height property instead.

```
.modal-content {
    max-height: calc(100vh - 200px);
    overflow-y: auto;
}
```

src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
```diff
@@ -5,6 +5,30 @@
         display: flex;
         flex-direction: column;
         gap: 0.375rem;
+
+        .singleselect-options {
+            max-height: 120px;
+            overflow-y: auto;
+            scrollbar-color: #a4a49b transparent;
```
mehalyna: The scrollbar styling lacks overflow-x: hidden on the .singleselect-options element, which could allow horizontal scrolling if content exceeds the width. Add overflow-x: hidden to prevent unwanted horizontal scroll.

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
```diff
@@ -13,4 +13,35 @@
     display: flex;
     flex-direction: column;
     gap: 0.4375rem;
+    overflow-y: visible;
+
+    .singleselect-options {
+        position: absolute;
```
mehalyna: The absolute positioning of .singleselect-options without a parent container having position: relative can cause it to position relative to the nearest positioned ancestor, leading to unpredictable placement. Ensure the parent container has position: relative.

mehalyna (commented):
Inconsistent scrollbar styling between the two modals: TeamCategoryModal.scss uses scrollbar-color but lacks the position: absolute and z-index properties present in DeleteTeamCategoryModal.scss, which could cause dropdown cropping in this modal too. Add positioning properties for consistency.

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
momaksymenko: Already fixed!

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
```diff
@@ -13,4 +16,42 @@
...
+        position: relative;
+
+        &:has(.singleselect-options) {
+            margin-bottom: calc(#{$dropdown-max-height} + #{$dropdown-gap}); 
+        }
+    }
+
+    .singleselect-options {
+        position: absolute;
+        top: 100%;
```
fhyhtyh: This overrides the shared top: calc(100% + 2px) from SingleSelectInput.scss, so this dropdown sits flush against the input while every other select in the app keeps a 2px gap. Intentional, or should we drop the override?

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
```diff
@@ -13,4 +16,42 @@
     display: flex;
     flex-direction: column;
     gap: 0.4375rem;
+    overflow-y: hidden;
+
+    .input-group {
+        position: relative;
+
```
fhyhtyh: input-group.scss:2 already sets position: relative on .input-group globally — this re-declaration is a no-op.

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
```diff
@@ -13,4 +16,42 @@
...
+        &:has(.singleselect-options) {
+            margin-bottom: calc(#{$dropdown-max-height} + #{$dropdown-gap}); 
+        }
+    }
+
+    .singleselect-options {
+        position: absolute;
+        top: 100%;
+        left: 0;
```
fhyhtyh: SingleSelectInput.scss already sets position: absolute; left: 0; on .singleselect-options — redundant.

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
```diff
@@ -13,4 +16,42 @@
...
+        right: 0;
```
fhyhtyh: This is inert: the base rule already sets width: 100% on .singleselect-options, so with left: 0 also set the box is over-constrained and right gets ignored.

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
```diff
@@ -13,4 +16,42 @@
...
+        z-index: 100;
+        max-height: $dropdown-max-height;
+        overflow-y: auto;
+        scrollbar-color: #a4a49b transparent;
```
fhyhtyh: #a4a49b is exactly colors.$grey-700, and this file already @uses colors (line 1). Can we use colors.$grey-700 here?

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
```diff
@@ -13,4 +16,42 @@
...
+            width: 8px;
+        }
+
+        &::-webkit-scrollbar-track {
+            background: transparent;
+            border-radius: 8px;
+        }
+
+        &::-webkit-scrollbar-thumb {
+            background-color: #a4a49b;
```
fhyhtyh: Same hardcoded #a4a49b — colors.$grey-700.

src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
```diff
@@ -5,6 +5,40 @@
         display: flex;
         flex-direction: column;
         gap: 0.375rem;
+
+        .input-group {
+            position: relative;
+        }
+
```
fhyhtyh: input-group.scss:2 already sets position: relative on .input-group globally — this re-declaration is a no-op.

src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
```diff
@@ -5,6 +5,40 @@
...
+        .singleselect-options {
+            position: absolute;
+            top: 100%;
```
fhyhtyh: Overrides the shared top: calc(100% + 2px) from SingleSelectInput.scss — same concern as the delete-category modal, this dropdown loses the 2px gap every other select has.

src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
```diff
@@ -5,6 +5,40 @@
...
+            left: 0;
```
fhyhtyh: Already set by the base .singleselect-options rule in SingleSelectInput.scss — redundant.

src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
```diff
@@ -5,6 +5,40 @@
...
+            right: 0;
```
fhyhtyh: Inert: width: 100% is already set on the base rule, so combined with left: 0 the box is over-constrained and right is ignored.

src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
```diff
@@ -5,6 +5,40 @@
...
+            z-index: 100;
+            max-height: 120px;
+            overflow-y: auto;
+            overflow-x: hidden;
+            scrollbar-color: #a4a49b transparent;
```
fhyhtyh: #a4a49b is exactly colors.$grey-700, and this file already @uses colors.

src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
```diff
@@ -5,6 +5,40 @@
...
+                width: 8px;
+            }
+
+            &::-webkit-scrollbar-track {
+                background: transparent;
+                border-radius: 8px;
+            }
+
+            &::-webkit-scrollbar-thumb {
+                background-color: #a4a49b;
```
fhyhtyh: Same hardcoded #a4a49b — colors.$grey-700.

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
momaksymenko: Okay, I will fix it

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
momaksymenko: Thank you, I will change the color. I used #a4a49b because I saw that it is used in other dropdown lists

mehalyna (approved):

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
```diff
@@ -13,4 +16,36 @@
...
+            border-radius: 8px;
+        }
+
+        &::-webkit-scrollbar-thumb {
+            background-color: colors.$grey-700;
+            border-radius: 8px;
+
+            &:hover {
+                background-color: colors.$grey-600;
+            }
```
fhyhtyh: colors.$grey-600 is lighter than the base colors.$grey-700 (line 43), so this lightens the thumb on hover instead of darkening it. Likely should be colors.$grey-800.

src/pages/admin/team/components/team-category-modal/TeamCategoryModal.scss
```diff
@@ -5,6 +5,32 @@
...
+                background: transparent;
+                border-radius: 8px;
+            }
+
+            &::-webkit-scrollbar-thumb {
+                background-color: colors.$grey-700;
+                border-radius: 8px;
+
+                &:hover {
+                    background-color: colors.$grey-600;
```
fhyhtyh: Same as DeleteTeamCategoryModal.scss:47 — colors.$grey-600 is lighter than the base colors.$grey-700 (line 26).

src/pages/admin/team/components/delete-team-category-modal/DeleteTeamCategoryModal.scss
momaksymenko: I didn't notice, thank you, I will fix it now

fhyhtyh (approved):

mehalyna:
## Solution for Failing Job 98519070368

The test failures are caused by **unhandled asynchronous state updates** in the `useDataPaginationFetch` hook. React's testing library is detecting that state updates (`setError` on line 105 and `setIsLoading` on line 108) are occurring outside of the `act()` function, which is required when testing components with async operations.

### Root Cause

The hook performs async operations in `fetchData()` that update state after promises resolve. When tests trigger these operations but don't wrap the async behavior in `act()`, React warns about updates not being wrapped properly.

### Solution

Wrap the state updates that occur after async operations complete. Here's the fixed version:

```typescript
const fetchData = useCallback(
    async (isReset: boolean) => {
        abortControllerRef.current?.abort();
        const newAbortController = new AbortController();
        abortControllerRef.current = newAbortController;

        setIsLoading(true);
        setError(null);
        if (isReset) {
            setData([]);
            currentPageRef.current = 0;
        }

        try {
            const pageToFetch = isReset ? 0 : currentPageRef.current;
            const paginationParams: PaginationRequestParams = {
                offset: pageToFetch * pageSize,
                limit: pageSize,
                requestOptions: { cancellationSignal: newAbortController.signal },
            };

            const result = await fetchHandler(paginationParams);

            // Check abort status before updating state
            if (newAbortController.signal.aborted) return;

            const totalFetchedPages = pageToFetch + 1;
            const maxPossibleItems = totalFetchedPages * pageSize;
            const newHasMore = maxPossibleItems < result.totalItemsCount;

            setData((prev) => {
                if (isReset) {
                    return result.items;
                }

                if (getUniqueId) {
                    const existingIds = new Set(prev.map((item) => getUniqueId(item)));
                    const uniqueItems = result.items.filter((item) => !existingIds.has(getUniqueId(item)));
                    return [...prev, ...uniqueItems];
                }

                return [...prev, ...result.items];
            });

            setHasMore(newHasMore);
            currentPageRef.current = pageToFetch + 1;
        } catch (error: any) {
            if (axios.isCancel?.(error) || error.name === 'CanceledError' || error.name === 'AbortError') {
                return;
            }
            // Check abort status before setting error
            if (!newAbortController.signal.aborted) {
                setError(error);
            }
        } finally {
            if (!newAbortController.signal.aborted) {
                setIsLoading(false);
            }
        }
    },
    [getUniqueId, pageSize, fetchHandler],
);
```

### Key Changes

1. **In the catch block (line 105)**: Add a check for `!newAbortController.signal.aborted` before calling `setError(error)` to prevent state updates on unmounted or cancelled requests.

2. **Already present (line 107-108)**: The finally block correctly checks abort status before `setIsLoading(false)`, but ensure this pattern is consistent.

This prevents state updates from being scheduled after the component unmounts or the request is cancelled, which eliminates the "not wrapped in act(...)" warnings during testing.

maromexe (approved):

PR #578, opened 2026-08-25
Feature/issue 2248 [Hippotherapy] View Scientific references

Zest005:
## Github ticket

* [Github ticket](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2248)

## Description

The Scientific references section rendered every reference fully expanded, and the expand button was a non-functional placeholder left by the layout story #3316. This PR implements the collapsed/expanded behaviour described in the user story.

## How it looks

image

## Summary of change

- `ScientificReferenceCard.tsx` - the expand button is now functional: it toggles the card through the new isExpanded / onToggleExpand props, swaps the arrow icon for a cross, and updates its aria-label between "Expand reference" and "Collapse reference". The collapsed card keeps the "Назва" field visible and hides only "Посилання".
- `ScientificReferencesSection.tsx` - holds the expanded state as a Set of localIds. A set rather than a single id, so several references can stay expanded at the same time, which is what an admin needs when comparing sources. On load the set is empty, so every entry starts collapsed.
- Tests - two new cases in ScientificReferenceCard.test.tsx for the collapsed and expanded rendering plus one for the toggle callback; three new cases in ScientificReferencesSection.test.tsx covering the default collapsed state, toggling, and keeping several entries expanded. Existing tests that asserted on the "Посилання" field were updated to render the card expanded.

## How to recreate changes

1. Log in to the admin panel and open /admin-panel/hippotherapy
2. Scroll to the "Наукові дослідження" section
3. All references are collapsed - only the "Назва" field is visible on each card
4. Click the arrow button on a card - the "Посилання" field appears and the icon changes to a cross
5. Click the arrow on another card - both stay expanded
6. Click the cross - the card collapses again

## CHECK LIST

### DIFF
--- a/src/pages/admin/hippotherapy-page/components/sections/scientific-references-section/scientific-reference-card/ScientificReferenceCard.tsx
+++ b/src/pages/admin/hippotherapy-page/components/sections/scientific-references-section/scientific-reference-card/ScientificReferenceCard.tsx
@@ -4,6 +4,7 @@ import { InputWithCharacterLimitGroup } from '@/components/admin/input-groups/in
 import { IconButton } from '@/components/admin/icon-button/IconButton';
 import { ACTION_ICONS } from '@/const/common/action-icons';
 import { ReactComponent as ArrowIcon } from '@/assets/icons/arrow-up-right.svg';
+import { ReactComponent as CrossIcon } from '@/assets/icons/cross.svg';
 import { HIPPOTHERAPY_PAGE_CHAR_LIMITS, HIPPOTHERAPY_PAGE_TEXT } from '@/const/admin/hippotherapy-page';
 import './ScientificReferenceCard.scss';
 
@@ -16,6 +17,8 @@ export interface ScientificReferenceCardProps {
     nameError?: string;
     urlError?: string;
     disabled?: boolean;
+    isExpanded: boolean;
+    onToggleExpand: (localId: string) => void;
     onNameChange: (localId: string, value: string) => void;
     onUrlChange: (localId: string, value: string) => void;
     onNameBlur: (localId: string) => void;
@@ -31,6 +34,8 @@ export const ScientificReferenceCard = ({
     nameError,
     urlError,
     disabled,
+    isExpanded,
+    onToggleExpand,
     onNameChange,
     onUrlChange,
     onNameBlur,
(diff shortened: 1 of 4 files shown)


src/pages/admin/hippotherapy-page/components/sections/scientific-references-section/scientific-reference-card/ScientificReferenceCard.tsx
```diff
@@ -85,18 +91,20 @@ export const ScientificReferenceCard = ({
...
+                        id={`hippotherapy-research-citation-${localId}-url`}
+                        name={`hippotherapy-research-citation-${localId}-url`}
+                        value={url}
+                        onChange={(event) => onUrlChange(localId, event.target.value)}
+                        onBlur={() => onUrlBlur(localId)}
+                        maxLength={HIPPOTHERAPY_PAGE_CHAR_LIMITS.RESEARCH_CITATION_URL}
+                        error={urlError}
+                        disabled={disabled}
+                    />
+                )}
```
mehalyna, line 108: Before this PR the empty URL input and its error were always on screen. Now the card is collapsed by default and the input is unmounted, and urlError renders inside the conditional block - so an admin filling in a fresh page hits a permanently disabled "Publish" button with no on-screen indication of what is missing. The same applies to any existing page whose URL gets cleared and the card collapsed.

Either surface invalid state on the collapsed header (error marker / auto-expand cards with a non-empty urlError), or keep the field mounted and hide it via CSS so its error remains reachable. Whichever the design allows.

url is required and part of the page-level validity gate:
- HippotherapyPageSchema.ts (lines 55-57) - scientificReferences.every(r => isValidText(r.name) && isValidText(r.url))
- HippotherapyPageContent.tsx (line 378) - disabled={!isPublishButtonActive || !isFormValid || ...}
- Blank-page default is one reference with url: '' (HippotherapyPageContent.tsx line 109)

mehalyna (commented):
localId is regenerated on every read: hippotherapy-page-api.ts line 45, called from both get and update (line 104), and handlePublish writes the result back into the draft (HippotherapyPageContent.tsx line 244). Since expandedIds is keyed by localId, every successful save collapses all cards. Old ids also stay in the Set forever (harmless in size, but it is stale state). 
Key expansion off a stable identity (id when present), or reconcile expandedIds against the current reference list.

src/pages/admin/hippotherapy-page/components/sections/scientific-references-section/scientific-reference-card/ScientificReferenceCard.tsx
Zest005: Thanks for comments, fixed. Expansion is keyed off id now, and the url error shows on the collapsed card too.
Surfacing why "Publish" is disabled on a blank, never-filled page belongs to the issue #2258 - on a fresh page urlError is never set, since errors only appear on blur. I will fix it there.

mehalyna (approved):

Mycola23 (approved):
In general looks pretty good especially after your fixes

PR #580, opened 2026-08-25
Fix/issue-3445 [Team] [Admin Panel] 'Назва' and 'Опис' fields in the 'Редагувати категорію' do not trim or collapse multiple consecutive spaces on edit

Valience:
## Github ticker

* [3445](https://github.com/ita-social-projects/VictoryCenter-Back/issues/3445)

## Description

Adds client-side validation preventing the 'Назва' and 'Опис' fields in the Team Category edit
modal from being saved with leading/trailing spaces. 

## How it looks

{69043ED5-0BC9-4939-975A-AF62139613E3}
{E342F102-6EF6-4A02-B89A-9932C11BACD3}

## Summary of change

- **`team.ts`**: Added `getNoSpacesError()` messages for `name` and `description`.
- **`team-category-schema.ts`**: Added a `no-leading-trailing-spaces` Yup `.test()` to both fields.
- **`team-category-schema.test.ts`**: Added tests covering leading-space and trailing-space cases
  for both `name` and `description`.

## How to recreate changes

1. Open 'Редагувати категорію' modal in the admin panel.
2. Enter a name or description with a leading or trailing space.
3. Observe the validation error preventing the space from being saved.

## CHECK LIST
- [x] New tests was added or existing was modified
- [ ] Changes has severe impact on users
- [ ] Changes has medium impact on users
- [x] Changes has light impact on users
- [x] Test covetage was updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Team category names and descriptions now reject leading or trailing spaces.
  * Invalid spacing displays a clear validation message.
  * Input is no longer silently trimmed, helping users correct formatting themselves.

### DIFF
--- a/src/const/admin/team.ts
+++ b/src/const/admin/team.ts
@@ -113,6 +113,7 @@ export const TEAM_CATEGORY_VALIDATION = {
         getMinError: () => `Не менше ${TEAM_CATEGORY_VALIDATION.name.min} символів`,
         getMaxError: () => `Не більше ${TEAM_CATEGORY_VALIDATION.name.max} символів`,
         getDuplicateNameError: () => COMMON_TEXT_ADMIN.CATEGORIES.FORM.MESSAGE.ALREADY_CONTAIN_CATEGORY_WITH_NAME,
+        getNoSpacesError: () => 'Назва не повинна містити пробіли на початку або в кінці',
     },
     description: {
         min: 5,
@@ -120,6 +121,7 @@ export const TEAM_CATEGORY_VALIDATION = {
         getRequiredError: () => 'Опис обов’язковий',
         getMinError: () => `Не менше ${TEAM_CATEGORY_VALIDATION.description.min} символів`,
         getMaxError: () => `Не більше ${TEAM_CATEGORY_VALIDATION.description.max} символів`,
+        getNoSpacesError: () => 'Опис не повинен містити пробіли на початку або в кінці',
     },
     teamMembersCount: {
         getHasTeamMembersCountError: (count: number) => `Категорія містить ${count} членів команди`,
--- a/src/validation/admin/team-category-schema/team-category-schema.test.ts
+++ b/src/validation/admin/team-category-schema/team-category-schema.test.ts
@@ -5,10 +5,14 @@ describe('TeamCategoryValidationSchema', () => {
     const validName = 'N'.repeat(TEAM_CATEGORY_VALIDATION.name.min + 1);
     const shortName = 'N'.repeat(TEAM_CATEGORY_VALIDATION.name.min - 1);
     const longName = 'N'.repeat(TEAM_CATEGORY_VALIDATION.name.max + 1);
+    const leadingSpaceName = ` ${validName}`;
+    const trailingSpaceName = `${validName} `;
 
     const validDescription = 'D'.repeat(TEAM_CATEGORY_VALIDATION.description.min + 1);
     const shortDescription = 'D'.repeat(TEAM_CATEGORY_VALIDATION.description.min - 1);
     const longDescription = 'D'.repeat(TEAM_CATEGORY_VALIDATION.description.max + 1);
+    const leadingSpaceDescription = ` ${validDescription}`;
+    const trailingSpaceDescription = `${validDescription} `;
 
     describe('name validation', () => {
         it('passes with valid name', async () => {
@@ -42,6 +46,18 @@ describe('TeamCategoryValidationSchema', () => {
         it('fails when name is too long', async () => {
             expect(TEAM_CATEGORY_VALIDATION_FUNCTIONS.validateName(longName)).toBe(
                 TEAM_CATEGORY_VALIDATION.name.getMaxError(),
+            );
+        });
+
+        it('fails when name has a leading space', async () => {
+            expect(TEAM_CATEGORY_VALIDATION_FUNCTIONS.validateName(leadingSpaceName)).toBe(
+                TEAM_CATEGORY_VALIDATION.name.getNoSpacesError(),
+            );
+        });
+
+        it('fails when name has a trailing space', async () => {
+            expect(TEAM_CATEGORY_VALIDATION_FUNCTIONS.validateName(trailingSpaceName)).toBe(
+                TEAM_CATEGORY_VALIDATION.name.getNoSpacesError(),
             );
         });
     });
@@ -78,6 +94,18 @@ describe('TeamCategoryValidationSchema', () => {
         it('fails when description is too long', async () => {
             expect(TEAM_CATEGORY_VALIDATION_FUNCTIONS.validateDescription(longDescription)).toBe(
                 TEAM_CATEGORY_VALIDATION.description.getMaxError(),
+            );
+        });
+        
+        it('fails when description has a leading space', async () => {
+            expect(TEAM_CATEGORY_VALIDATION_FUNCTIONS.validateDescription(leadingSpaceDescription)).toBe(
+                TEAM_CATEGORY_VALIDATION.description.getNoSpacesError(),
+            );
+        });
+
+        it('fails when description has a trailing space', async () => {
+            expect(TEAM_CATEGORY_VALIDATION_FUNCTIONS.validateDescription(trailingSpaceDescription)).toBe(
+                TEAM_CATEGORY_VALIDATION.description.getNoSpacesError(),
             );
         });
     });
--- a/src/validation/admin/team-category-schema/team-category-schema.ts
+++ b/src/validation/admin/team-category-schema/team-category-schema.ts
@@ -3,14 +3,22 @@ import * as Yup from 'yup';
 
 export const TeamCategoryValidationSchema = Yup.object({
     name: Yup.string()
-        .trim()
         .required(TEAM_CATEGORY_VALIDATION.name.getRequiredError())
+        .test(
+            'no-leading-trailing-spaces',
+            TEAM_CATEGORY_VALIDATION.name.getNoSpacesError(),
+            (value) => value === value?.trim(),
+        )
         .min(TEAM_CATEGORY_VALIDATION.name.min, TEAM_CATEGORY_VALIDATION.name.getMinError())
         .max(TEAM_CATEGORY_VALIDATION.name.max, TEAM_CATEGORY_VALIDATION.name.getMaxError()),
 
     description: Yup.string()
-        .trim()
         .required(TEAM_CATEGORY_VALIDATION.description.getRequiredError())
+        .test(
+            'no-leading-trailing-spaces',
+            TEAM_CATEGORY_VALIDATION.description.getNoSpacesError(),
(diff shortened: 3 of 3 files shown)


src/validation/admin/team-category-schema/team-category-schema.ts
```diff
@@ -3,14 +3,22 @@ import * as Yup from 'yup';
 
 export const TeamCategoryValidationSchema = Yup.object({
     name: Yup.string()
-        .trim()
         .required(TEAM_CATEGORY_VALIDATION.name.getRequiredError())
+        .test(
+            'no-leading-trailing-spaces',
+            TEAM_CATEGORY_VALIDATION.name.getNoSpacesError(),
+            (value) => value === value?.trim(),
+        )
```
mehalyna: The custom .test() validation runs after .required(), but .required() will fail on empty strings before whitespace validation occurs. Consider ordering the test before .required() for consistent error messaging when users clear and try to save.

src/validation/admin/team-category-schema/team-category-schema.ts
```diff
@@ -3,14 +3,22 @@ import * as Yup from 'yup';
...
         .max(TEAM_CATEGORY_VALIDATION.name.max, TEAM_CATEGORY_VALIDATION.name.getMaxError()),
 
     description: Yup.string()
-        .trim()
         .required(TEAM_CATEGORY_VALIDATION.description.getRequiredError())
+        .test(
+            'no-leading-trailing-spaces',
+            TEAM_CATEGORY_VALIDATION.description.getNoSpacesError(),
+            (value) => value === value?.trim(),
+        )
```
mehalyna: Same ordering issue as the name field—the whitespace test should execute before .required() to provide the correct error message priority.

src/validation/admin/team-category-schema/team-category-schema.test.ts
```diff
@@ -44,6 +48,18 @@ describe('TeamCategoryValidationSchema', () => {
                 TEAM_CATEGORY_VALIDATION.name.getMaxError(),
             );
         });
+
+        it('fails when name has a leading space', async () => {
+            expect(TEAM_CATEGORY_VALIDATION_FUNCTIONS.validateName(leadingSpaceName)).toBe(
+                TEAM_CATEGORY_VALIDATION.name.getNoSpacesError(),
+            );
+        });
```
mehalyna, line 56: The test expects validateName() to return an error for leading/trailing spaces, but the validation function does not trim before validation in the test context. Ensure the test data matches actual field behavior - consider trimming test inputs or verifying the exact error message returned.

src/validation/admin/team-category-schema/team-category-schema.ts
```diff
@@ -3,14 +3,22 @@ import * as Yup from 'yup';
 
 export const TeamCategoryValidationSchema = Yup.object({
     name: Yup.string()
-        .trim()
         .required(TEAM_CATEGORY_VALIDATION.name.getRequiredError())
+        .test(
+            'no-leading-trailing-spaces',
+            TEAM_CATEGORY_VALIDATION.name.getNoSpacesError(),
+            (value) => value === value?.trim(),
```
fhyhtyh: Since .trim() is gone, any category name saved before this fix with padding whitespace (the old schema validated the trimmed value but stored the raw one) will now fail validation the moment it's loaded for editing — and since useFormManager validates on load, the Save button stays disabled even for unrelated edits, with no way to fix it except retyping the field. Might be worth a one-time trim on load/normalize existing data, or trimming on blur before validating.

src/validation/admin/team-category-schema/team-category-schema.ts
```diff
@@ -3,14 +3,22 @@ import * as Yup from 'yup';
 
 export const TeamCategoryValidationSchema = Yup.object({
     name: Yup.string()
-        .trim()
         .required(TEAM_CATEGORY_VALIDATION.name.getRequiredError())
+        .test(
+            'no-leading-trailing-spaces',
```
fhyhtyh: This runs after .required(), so a whitespace-only value ("   ") passes required() (not empty/null) and then fails here instead — changing the error from "required" to "no leading/trailing spaces". This actually breaks the existing tests fails when name is only whitespace and fails when description is only whitespace (line ~35 / ~83 in the test file) — ran the suite locally and both are red on this branch.

src/validation/admin/team-category-schema/team-category-schema.ts
```diff
@@ -3,14 +3,22 @@ import * as Yup from 'yup';
...
         .min(TEAM_CATEGORY_VALIDATION.name.min, TEAM_CATEGORY_VALIDATION.name.getMinError())
         .max(TEAM_CATEGORY_VALIDATION.name.max, TEAM_CATEGORY_VALIDATION.name.getMaxError()),
 
     description: Yup.string()
-        .trim()
         .required(TEAM_CATEGORY_VALIDATION.description.getRequiredError())
+        .test(
+            'no-leading-trailing-spaces',
+            TEAM_CATEGORY_VALIDATION.description.getNoSpacesError(),
+            (value) => value === value?.trim(),
```
fhyhtyh: The no-leading-trailing-spaces test (id, message, predicate) is duplicated for name and description. Consider extracting a small helper (e.g. noLeadingTrailingSpacesTest(getMessage)) so future tweaks only need one update.

mehalyna:
## Solution

The ESLint/Prettier job failed due to **trailing whitespace** on line 99 of the test file.

**Issue:** Line 99 contains extra spaces at the end that need to be removed:

```typescript
        });
         // ← Line 99 has trailing whitespace here
        it('fails when description has a leading space', async () => {
```

**Fix:** Remove the trailing spaces on line 99. The corrected file should look like:

```typescript
        });
        it('fails when description has a leading space', async () => {
```

You can run `eslint --fix` to automatically correct this formatting issue, or simply delete the extra spaces at the end of line 99 in your editor.

**Source file:** [team-category-schema.test.ts](https://github.com/ita-social-projects/VictoryCenter-Client/blob/47009a88a1ffc500190251625d01d7d734135e16/src/validation/admin/team-category-schema/team-category-schema.test.ts)

mehalyna:
## Solution

The test is failing because **whitespace-only strings are not being properly validated**. The issue is in the validation schema logic.

**Problem:** When a string contains only whitespace (e.g., `'   '`), the `.trim()` method removes all characters, resulting in an empty string. The current schema checks:
1. `.required()` - passes because the original value is not empty
2. `.test('no-leading-trailing-spaces')` - passes because `'   ' === '   '.trim()` is `false` (correctly fails this test)
3. But the test expects the required error, not the spaces error

The validation order needs adjustment. The `.required()` test should handle both truly empty strings AND whitespace-only strings.

**Fix:** Update the validation schema to trim the value before checking if it's required:

```typescript
export const TeamCategoryValidationSchema = Yup.object({
    name: Yup.string()
        .required(TEAM_CATEGORY_VALIDATION.name.getRequiredError())
        .trim() // Add this to trim before validation
        .test(
            'no-leading-trailing-spaces',
            TEAM_CATEGORY_VALIDATION.name.getNoSpacesError(),
            (value) => value === value?.trim(),
        )
        .min(TEAM_CATEGORY_VALIDATION.name.min, TEAM_CATEGORY_VALIDATION.name.getMinError())
        .max(TEAM_CATEGORY_VALIDATION.name.max, TEAM_CATEGORY_VALIDATION.name.getMaxError()),

    description: Yup.string()
        .required(TEAM_CATEGORY_VALIDATION.description.getRequiredError())
        .trim() // Add this to trim before validation
        .test(
            'no-leading-trailing-spaces',
            TEAM_CATEGORY_VALIDATION.description.getNoSpacesError(),
            (value) => value === value?.trim(),
        )
        .min(TEAM_CATEGORY_VALIDATION.description.min, TEAM_CATEGORY_VALIDATION.description.getMinError())
        .max(TEAM_CATEGORY_VALIDATION.description.max, TEAM_CATEGORY_VALIDATION.description.getMaxError()),
});
```

Adding `.trim()` after `.required()` ensures that whitespace-only strings are treated as empty and trigger the "required" error message, which is what the test expects.

**Source files:**
- [team-category-schema.ts](https://github.com/ita-social-projects/VictoryCenter-Client/blob/47009a88a1ffc500190251625d01d7d734135e16/src/validation/admin/team-category-schema/team-category-schema.ts)
- [team-category-schema.test.ts](https://github.com/ita-social-projects/VictoryCenter-Client/blob/47009a88a1ffc500190251625d01d7d734135e16/src/validation/admin/team-category-schema/team-category-schema.test.ts)

fhyhtyh (commented):
Comments 2 and 3 were addressed, but the first one wasn't — the schema still has no trim and nothing normalizes on load, so padded existing data still locks the Save button with no recovery. Could you handle it with a trim-on-blur (and/or normalizing initialData on load), or confirm here if you think pre-existing padded rows aren't a real possibility?

src/const/admin/team.ts
```diff
@@ -113,13 +113,15 @@ export const TEAM_CATEGORY_VALIDATION = {
         getMinError: () => `Не менше ${TEAM_CATEGORY_VALIDATION.name.min} символів`,
         getMaxError: () => `Не більше ${TEAM_CATEGORY_VALIDATION.name.max} символів`,
         getDuplicateNameError: () => COMMON_TEXT_ADMIN.CATEGORIES.FORM.MESSAGE.ALREADY_CONTAIN_CATEGORY_WITH_NAME,
+        getNoSpacesError: () => 'Назва не повинна містити пробіли на початку або в кінці',
```
Denys1310, line 116: Since this field is normalized when we move the focus away, is it necessary to show this message when the value has already been fixed? Consider it for description field as well

Image

Denys1310 (commented):
Unfortunately, the issue described in the assigned ticket is not reproducible. After saving, the value still contains **consecutive spaces**. Please make sure **consecutive spaces** are handled here, or check whether the assigned ticket is valid here

Image

mehalyna (approved):

Denys1310 (approved):
Thank you for fix. Please make sure all checks are green before merge

fhyhtyh (commented):
The same normalization fix landed in TeamCategoryModal, but TranslateTeamCategoryForm / TranslateTeamCategoryModal use the same schema and aren't covered here. Since this PR is what drops the schema-level .trim(), editing a localization whose name/description was stored with leading/trailing whitespace (possible for rows created via the translate form before this change, when the old non-strict .trim() accepted padding) will now fail validation on load — and because useFormManager validates on mount, the Save button stays disabled with no way to recover, since the blur handlers here validate the raw value without normalizing.

Could we apply the same two changes to the translate path?

* trim existingLocalization.name / .description in the initialData useMemo (TranslateTeamCategoryModal.tsx:54-57)
* normalize (trim().replace(/\s+/g, ' ')) before validating in handleNameBlur / handleDescriptionBlur (TranslateTeamCategoryForm.tsx:96-108), mirroring TeamCategoryModal

PR #581, opened 2026-08-26
Fix/issue-2249-2250-2251[Hippotherapy] [Admin Panel] Align Hippotherapy text fields with user stories #2249 #2250 #2251

maromexe:
## JIRA or Github ticker

* [[JIRA or Github ticket](https://jira.softserve.academy/secure/RapidBoard.jspa?rapidView=id)](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2249)
* [[JIRA or Github ticket](https://jira.softserve.academy/secure/RapidBoard.jspa?rapidView=id)](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2250)
* [[JIRA or Github ticket](https://jira.softserve.academy/secure/RapidBoard.jspa?rapidView=id)](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2251)

## Description

Updated the text fields on the Hippotherapy admin page according to the user stories and acceptance criteria.
**#2249 — Text field character limits**

- Applied the required character limits to Hippotherapy page text fields.
Added separate minimum lengths for titles and text fields:
Title — 5 characters
Text — 10 characters
- Updated optional fields according to the user story.
- Updated the quote author's name field to be optional with a 50-character limit.
- Updated the additional description field to be optional with a 50-character limit.

**#2250 — Text field validation**

- Changed text field validation from onChange to onBlur.
- After a validation error appears, the field is revalidated on change so the error disappears immediately after the value becomes valid.
- Updated validation helpers and schema to support different minimum lengths for titles and text fields.

**#2251 — Rich text input**

- Updated RichTextInput usage across all Hippotherapy page sections.
- Replaced the English ʼEnter text...ʼ placeholder with the Ukrainian ʼВведіть текст...ʼ.
- Updated related tests.

## How it looks

image

## Summary of change

- Aligned Hippotherapy text fields with the requirements from #2249, #2250 and #2251.
- Updated character limits and optional fields.
- Updated validation behavior and minimum-length rules.
- Updated RichTextInput placeholder.
- Updated and added tests for the affected functionality.
## How to recreate changes

1. Open the Hippotherapy page in the admin panel.
2. Open any text field.
3. Verify the character limit and validation behavior.
4. Leave a field with an invalid value and move focus away from it.
5. Verify that the validation message is displayed.
6. Correct the value and verify that the validation message disappears.
7. Verify that RichTextInput displays the Ukrainian Введіть текст... placeholder.

## CHECK LIST

## Summary by CodeRabbit

* **New Features**
  * Added consistent rich-text validation across hippotherapy administration fields.
  * Validation errors now appear on blur and clear as users edit.
  * Added dedicated handling for image updates and image errors.
  * Added reusable gallery card and ethics principle components.
* **Bug Fixes**
  * Applied field-specific minimum and maximum text lengths.
  * Made applicable description and quote author fields optional.
  * Updated the default rich-text placeholder to “Введіть текст...”.
* **Tests**
  * Expanded coverage for validation, image handling, blur behavior, and disabled fields.


src/hooks/admin/use-hippotherapy-section-fields/useHippotherapySectionFields.ts
Denys1310: According to the requirements, the 'Зберегти' button should be enabled when all validation checks pass successfully. However, it is currently disabled. If this is caused by another change or dependency, please make sure the button works correctly together with the validation logic

Image

maromexe: Looked into this. The button is disabled because the image checks fail - 
isValidImage requires an uploaded image or an existing imageId for the intro, 
quote, hippovention-center, another-quote and ethics banners, plus all 8 
gallery cards. That check comes from #569 and isn't touched in this PR

src/pages/admin/hippotherapy-page/components/sections/gallery-section/GallerySection.tsx
```diff
@@ -70,6 +104,7 @@ const GallerySectionComponent = ({
                                 variant="whoWeAre"
                                 value={card.image}
                                 onChange={(image) => handleCardImageChange(index, image)}
+                                onBlur={() => handleCardDescriptionBlur(index)}
```
mehalyna: …while the description field itself (lines 125-135) has onChange but no onBlur. It is the only rich-text field in the whole page without one.

Two defects follow:
1. Tabbing out of the image input surfaces an error under the description field the user never touched.
2. Blurring the description itself never validates. Because handleCardDescriptionChange re-validates only if (cardDescriptionErrors[index] !== undefined), the error can never appear through normal interaction with that field.
Impact is amplified by the fact that publish is gated on isHippotherapyPageContentValid (HippotherapyPageContent.tsx line 175): a too-short card description silently disables Publish with no on-screen reason. GallerySection renders both the advantages and participants sections (lines 327, 358), so this affects 2 sections × N cards.

Move onBlur onto the description RichTextInputGroup; decide separately whether the image input needs a blur hook at all (it currently has none of its own semantics here).

src/const/admin/hippotherapy-page.ts
```diff
@@ -30,11 +31,11 @@ export const HIPPOTHERAPY_PAGE_CHAR_LIMITS = {
     INTRO_DESCRIPTION: 300,
     TEXT_CARD_TITLE: 50,
     TEXT_CARD_DESCRIPTION: 1000,
-    QUOTE_TEXT: 100,
+    QUOTE_TEXT: 300,
     QUOTE_AUTHOR: 50,
     HIPPOVENTION_CENTER_TITLE: 50,
     HIPPOVENTION_CENTER_PROS: 300,
-    HIPPOVENTION_CENTER_DESCRIPTION: 300,
+    HIPPOVENTION_CENTER_DESCRIPTION: 50,
```
mehalyna, line 40: MaxLengthPlugin does not truncate on load - it truncates on the first user edit, via a RootNode transform that calls trimTextContentFromAnchor(editor, selection.anchor, overflowLength) (MaxLengthPlugin.tsx lines 35-39). So an already-published 210-character description loads intact and displays 210/50, but as soon as the admin types a single character in that field, ~160 characters are removed from around the caret. If they then publish, the truncation persists.

Also note the resulting window is narrow: the field is optional, but once non-empty it must be ≥10 and ≤50 characters.

Confirm 50 is what the story specifies (it reads more like a title limit than a description limit); if it is, plan for existing over-limit content - e.g. a migration, or surfacing an explicit over-limit error instead of silent trimming.

src/hooks/admin/use-hippotherapy-section-fields/useHippotherapySectionFields.test.ts
```diff
@@ -85,4 +50,13 @@ describe('useHippotherapySectionFields', () => {
...
     });
+
+    it('exposes the text field handlers from useHippotherapyTextFields', () => {
+        const { result } = renderHook(() => useHippotherapySectionFields({ value: initialValue, onChange }));
+
+        expect(result.current.handleTitleChange).toBeDefined();
+        expect(result.current.handleTitleBlur).toBeDefined();
+        expect(result.current.handleDescriptionChange).toBeDefined();
+        expect(result.current.handleDescriptionBlur).toBeDefined();
+    });
```
mehalyna: Two behavioural tests (calls onChange with the updated title and sets the validation error, and the description equivalent) were deleted and replaced by:
```
expect(result.current.handleTitleChange).toBeDefined();
expect(result.current.handleTitleBlur).toBeDefined();
```
This asserts only that a spread happened. Moving the assertions into useHippotherapyTextFields.test.ts covers the hook, but nothing now verifies that useHippotherapySectionFields still forwards value/onChange/isDescriptionOptional correctly - exactly the wiring a future refactor would break.

Separately, three sections received new branching logic and no test updates:

- GallerySection.tsx - handleTitleBlur, - handleCardDescriptionBlur (the existing test file was not touched, which is why Finding 1 slipped through)
- HippotherapyQuoteSection.tsx: lines 29-254 - the new optional-author empty/non-empty branches, the most intricate new logic in the PR
- ScientificReferencesSection.tsx - hook adoption + MIN_TITLE_LENGTH switch
Given the repo enforces 93.5% lines / 86.9% branches, the added untested branches are a plausible threshold failure.

src/validation/admin/hippotherapy-page-schema/HippotherapyPageSchema.ts
```diff
@@ -4,65 +4,75 @@ import { HIPPOTHERAPY_PAGE_TEXT } from '@/const/admin/hippotherapy-page';
...
+        isValidTitle(content.analysisSection.title) &&
         isValidHtmlText(content.analysisSection.description) &&
-        isValidHtmlText(content.scientificReferencesSection.title) &&
+        isValidTitle(content.scientificReferencesSection.title) &&
         isValidHtmlText(content.scientificReferencesSection.description) &&
         content.scientificReferencesSection.scientificReferences.every(
-            (reference) => isValidText(reference.name) && isValidText(reference.url),
+            (reference) =>
+                isValidText(reference.name, HIPPOTHERAPY_PAGE_TEXT.MIN_TITLE_LENGTH) &&
+                isValidText(reference.url, HIPPOTHERAPY_PAGE_TEXT.MIN_TITLE_LENGTH),
```
mehalyna: A URL is not a title; reusing MIN_TITLE_LENGTH couples two unrelated rules, so a future title-length change silently moves URL validation with it. The practical effect is that the URL bar is now weaker than before (5 instead of 10) with still no Yup.string().url() or scheme check - abcde passes and gets published as a scientific reference link. Lack of format validation is pre-existing; the loosening is new.

mehalyna (commented):
The refactor commit introduced useHippotherapyTextFields but only TextCardSection, ScientificReferencesSection, and (indirectly) useHippotherapySectionFields adopted it. The same if (xError !== undefined) { revalidate } + handleXBlur pair is now duplicated in:
- GallerySection.tsx lines 37-82 (title + card description)
- HippotherapyQuoteSection.tsx lines 46-70 (quote text + author)
- HippoventionCenterSection.tsx lines 44-53 (pros)
- EthicsSection.tsx lines 38-52 (principles)
That divergence is the the hook guarantees a handleXBlur exists for every handleXChange; hand-rolled copies do not.

The shared abstraction wants to be per-field (useValidatedRichTextField(value, onChange, minLength, optional)) rather than per-section with a fixed {title, description} shape; that shape is precisely why the four sections above couldn't adopt it.

Denys1310 (approved):

src/pages/admin/hippotherapy-page/components/sections/gallery-section/GallerySection.tsx
maromexe: Fixed - moved onBlur from the ImageInput onto the description 
RichTextInputGroup. The card is now a separate GalleryCard component with its 
own useValidatedRichTextField, so change and blur are wired to the same field 
by construction

Added tests - one for the description blur, one asserting the image blur 
doesn't surface an error under the description

src/const/admin/hippotherapy-page.ts
maromexe: Confirmed 50 is per the story - the field table in #2249 lists "Додатковий 
опис" / "В центрі іповенції" as Optional, Char (50), and the Figma mockup 
shows it holding a person's name (50/50), so the limit fits the intended 
content

src/hooks/admin/use-hippotherapy-section-fields/useHippotherapySectionFields.test.ts
maromexe: Restored the behavioural tests in the hook test file - they now verify 
value/onChange forwarding and the optional flag rather than just checking that 
a spread happened

Added tests for useValidatedRichTextField, GalleryCard and EthicsPrinciple, 
and blur tests for the quote and references sections

src/validation/admin/hippotherapy-page-schema/HippotherapyPageSchema.ts
maromexe: split into MIN_REFERENCE_NAME_LENGTH and MIN_REFERENCE_URL_LENGTH so 
title changes don't move URL rules

the 5 char minimum itself comes from the validation table in #2250 
("Посилання" - Min length 5 characters). URL format validation isn't in scope 
there - a separate ticket?

src/const/admin/hippotherapy-page.ts
mehalyna: If 50 is real: three things, in this order
1. Fix MaxLengthPlugin - this is the actual defect, and it is field-agnostic.

MaxLengthPlugin.tsx lines 33-40 computes overflow against the absolute limit, so it deletes all pre-existing excess on the first keystroke. It should only reject what the current update added:
```
const currentLength = currentTextContent.length;
const addedLength = currentLength - prevTextContent.length;
const overflow = Math.min(currentLength - maxLength, addedLength);

if (overflow > 0) {
    trimTextContentFromAnchor(editor, selection.anchor, overflow);
}
```

Scenario (max 50) |	Today |	With clamp
| :--- | :---: | ---: |
Empty field, paste 100 |	keeps 50 |	keeps 50 (unchanged)
210-char legacy value, type 1 char |	deletes 161 |	rejects the keystroke, content intact
210-char legacy value, press Backspace |	no trim |	no trim - user can shrink freely
210-char legacy value, paste 40	| deletes 200 |	rejects the paste, content intact

Be aware this is not airtight: select-all + replace with 60 chars yields addedLength < 0, so 60 survives under a 50 limit. That gap is why (step 2) matters - the plugin is a UX affordance, not the invariant.

Worth a regression test: mount RichTextInput with initial content longer than maxLength, type one character, assert the content lost exactly zero pre-existing characters. That test doesn't exist today, which is why this went unnoticed.

2. Add max-length to the publish gate. isHippotherapyPageContentValid (HippotherapyPageSchema.ts lines 41-79) validates only minimums - the maximum is enforced purely by the editor's silent trimming. So a value that arrives from the API over the limit is invisible to validation and publishes as-is. With (step 1) in place, that becomes the expected state for grandfathered content, so it needs a real error rather than silence.

The wiring is cheap: RichTextInputGroup already tracks currentLength via onLengthChange (RichTextInputGroup.tsx lines 31,55) and InputErrorWithCharacterCounter already exists - this is surfacing state that's already computed, not new infrastructure.

3. Check the one record in staging/prod and shorten hippoventionCenterSection.description by hand if it exceeds 50. No script needed.

stkossman (commented):
Good job overall. However look at my comments:

### 1. Link in rich text editor
In [2251](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2251) it is said:
_I want basic formatting controls (bold, italic, links) for text fields, **so that** I can apply formatting to present the description._

There's no link present in _all_ text fields. Is there a business requirements misconception and links shouldn't be present in all text fields?

Image

### 2. UI-Design
There's a difference between mockup and actual UI. Judging by the mockup, the arrow should be on the right of the text field, not near the Delete button. Verify with BA (or say it is intentional) whether it is right position for the arrow. 

Actual:
Image

Mockup:
Image

---

Also resolve all conflicts before merge

mehalyna (approved):

src/components/admin/rich-text-input/RichTextInput.tsx
```diff
@@ -62,7 +63,7 @@ export const RichTextInput = ({
     maxLength,
     disabled = false,
     hideToolbar = false,
-    placeholder = 'Enter text...',
+    placeholder = COMMON_TEXT_ADMIN.INPUT.ENTER_TEXT_PLACEHOLDER,
```
stkossman, line 66: Changing the shared `RichTextInput` default placeholder here makes unrelated admin pages inherit the Hippotherapy-specific text. This PR should keep the change scoped to Hippotherapy field usages instead of altering the global default.

For instance, the Who We Are admin description field now inherits the Hippotherapy-specific placeholder change.

Is this intentional? Say if yes. If not - fix.

maromexe:
@stkossman 

**1. Links** -  #2251 does mention them. The toolbar is a shared 
component used on Partners, Who We Are and Main too, and none of them have a 
link button. It's kind of outside the scope here, needs separate ticket and I'll take it

**2. Arrow position** - comes from #2248, merged separately. Not part of my 
tickets, so I haven't touched it. Also separate ticket needed, could take it

**3. Conflicts** - resolved, merged release/1.0.0 into the branch.

src/components/admin/rich-text-input/RichTextInput.tsx
maromexe: Intentional. The previous default was the English `Enter text...` in an 
otherwise Ukrainian admin panel, so Who We Are, Partners and Main were all 
showing an untranslated placeholder. The new value is a generic 
"Введіть текст..." in COMMON_TEXT_ADMIN.INPUT - it's not Hippotherapy-specific, 
and any page can still override it via the placeholder prop.

Scoping it to Hippotherapy would mean passing the same prop to every rich text 
field on the page and leaving the English default in place elsewhere, which 
seemed worse

stkossman:
> @stkossman
> 
> **1. Links** - #2251 does mention them. The toolbar is a shared component used on Partners, Who We Are and Main too, and none of them have a link button. It's kind of outside the scope here, needs separate ticket and I'll take it
> 
> **2. Arrow position** - comes from #2248, merged separately. Not part of my tickets, so I haven't touched it. Also separate ticket needed, could take it
> 
> **3. Conflicts** - resolved, merged release/1.0.0 into the branch.

1. About links in #2251 - I reckon this should be discussed with BA what "links" means in our context. But I guess it's just ticket's error, no need to worry.
2. If arrow changes comes from #2248 - I guess separate bug ticket should be created and taken into work. Previously discussed with BA whether move arrow or keep it in place for now.

Approved, no changes needed :)

stkossman (approved):

PR #585, opened 2026-08-27
Fix/issue-3503-3570 The 'Сума UAH' field is not recalculated.

cirillad:
## Github ticket

- [#3503](https://github.com/ita-social-projects/VictoryCenter-Back/issues/3503)
- [#3570](https://github.com/ita-social-projects/VictoryCenter-Back/issues/3570)

## Description

In the admin panel's Reports section, editing the "Сума USD" field did not recalculate the corresponding "Сума UAH" field — only the opposite direction (UAH → USD) worked. This affected two separate entry points that share the same underlying recalculation logic:

- **#3503** — the "Програмні витрати" Add/Edit record modal: entering a USD amount showed a "Сума в USD не відповідає сумі в UAH" error instead of recalculating UAH, blocking save entirely.
- **#3570** — inline row editing in the "Доходи та витрати" table (both "Надходження" and "Витрата"): the same one-directional behavior.

## Summary of change

- `get-converted-amount.ts` — added a `direction` parameter (`'uahToUsd' | 'usdToUah'`, defaults to the existing behavior) so the conversion formula can run in either direction.
- `update-funds-amounts.ts` — added a symmetric branch for `field === 'amountUsd'` that recalculates `amountUah`, mirroring the existing UAH→USD branch. This is the core fix.
- `useAmountBlur.ts` (used by the Add/Edit modals) — removed the special-cased USD-only blur handler that skipped recalculation entirely; both fields now go through the same `updateFundsAmounts` call on blur, matching the pattern already used by `useFundsAmountEdit.ts` (inline table editing, which required no production change).
- `useProgramExpenseRecordForm.ts` / `useFundsExpendituresRecordForm.ts` — no code changes needed; both hooks already delegated to the two files above and now recalculate correctly.
- Updated/added unit tests across all affected files (`get-converted-amount`, `update-funds-amounts`, `useAmountBlur`, `useFundsAmountEdit`, `useProgramExpenseRecordForm`, `useFundsExpendituresRecordForm`) plus the UI-level tests (`AddProgramExpenseRecordModal`, `FundsExpendituresTable`) that exercised the old one-directional behavior as "correct."

## How to recreate changes

**#3503:**
1. Open Admin Panel → "Звітність" → "Звіт та аналітика" → "Програмні витрати" tab → "Додати запис".
2. Enter a value in "Сума USD" only (leave UAH untouched or vice versa).
3. Confirm "Сума UAH" recalculates automatically and the record saves without a mismatch error.

**#3570:**
1. Open the "Доходи та витрати" tab → enter edit mode on any "Надходження" or "Витрата" row.
2. Change "Сума USD".
3. Confirm "Сума UAH" recalculates on screen before saving, and the saved value persists after a page refresh.

## CHECK LIST

- [x] New tests were added or existing were modified
- [ ] Changes have severe impact on users
- [x] Changes have medium impact on users
- [ ] Changes have light impact on users

### DIFF
--- a/src/hooks/admin/use-amount-blur/useAmountBlur.ts
+++ b/src/hooks/admin/use-amount-blur/useAmountBlur.ts
@@ -2,10 +2,6 @@ import { useCallback, useState } from 'react';
 import { FUNDS_EXPENDITURES_TEXT } from '@/const/admin/reports';
 import { updateFundsAmounts } from '@/utils/functions/update-funds-amounts/update-funds-amounts';
 import { isUsdAmountMismatch } from '@/utils/functions/validate-usd-amount-mismatch/validate-usd-amount-mismatch';
-import {
-    normalizeFundsExpendituresAmountInput,
-    validateFundsExpendituresAmount,
-} from '@/validation/admin/reports-schema/funds-expenditures-record-schema/funds-expenditures-record-schema';
 
 export const getUsdMismatchMessage = (
     amountUah: string,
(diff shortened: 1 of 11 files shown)


src/hooks/admin/use-amount-blur/useAmountBlur.ts
```diff
@@ -23,40 +19,23 @@ export const useAmountBlur = (exchangeRate: string | null, customMismatchMessage
...
+
+                setUsdMismatchMessage(
+                    field === 'amountUsd' && !suppressMismatchCheck
+                        ? getUsdMismatchMessage(
+                              updated.amountUah,
+                              updated.amountUsd,
+                              exchangeRate,
+                              customMismatchMessage,
+                          )
+                        : undefined,
```
mehalyna: setUsdMismatchMessage is only invoked with a real message when field === 'amountUsd'; blurring amountUah always clears the mismatch message (undefined), even though updateFundsAmounts may have just recalculated amountUsd from amountUah. This mirrors the pre-existing behavior (the old code also unconditionally set undefined for the amountUah branch), so it's not a regression from this diff, but it means a UAH-blur that produces a mismatched (rounded) USD value won't surface the mismatch message until the USD field itself is blurred. Flagging for confirmation this is intended.

mehalyna (approved):

momaksymenko (approved):

src/hooks/admin/use-amount-blur/useAmountBlur.ts
cirillad: This was a leftover from the old one-directional logic. Made it symmetric: amountUah blur now runs the same mismatch check as amountUsd instead of hardcoding undefined. Added tests for it.

Denys1310 (approved):

PR #587, opened 2026-08-28
Feature/issue 2258 [Hippotherapy] Scientific references adding

Zest005:
## Github ticket

- [Github ticket](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2258)

## Description

The "Додати +" button was a placeholder. This PR implements adding a new scientific reference and applies space management to the "Назва" field.

## How it looks

https://github.com/user-attachments/assets/700ec538-9751-4f26-8b7a-f26b0ffe749e

## Summary of change

- `ScientificReferencesSection.tsx` - "Додати +" adds a new empty reference at the end of the list, expanded and focused. The button is disabled while any reference is incomplete and re-enabled once all are valid. Space management is applied to "Назва" on every keystroke.
- `text-formatters.ts` - added `getNormalizedInputTextWhileTyping`. It collapses consecutive spaces and strips leading ones, but keeps a trailing space.
- `TextAreaWithCharacterLimit.tsx` - required for the above. Unlike `InputWithCharacterLimit`, it never re-synced its internal localValue back to the value prop, so a parent could not correct what the user typed.
- Bug found while testing: blur validation ran against the untrimmed value, so 9 characters plus a space passed a minimum of 10. Both blur handlers now validate the trimmed value.
- "Опублікувати" needed no change - `isHippotherapyPageContentValid` already covers references.

Out of scope: validation rules themselves [#2250](https://github.com/ita-social-projects/VictoryCenter-Back/issues/2250). This section delegates to `HIPPOTHERAPY_PAGE_VALIDATION_FUNCTIONS.validateText`. Once it merges, these call sites will need the explicit length argument.

## How to recreate changes

1. Open /admin-panel/hippotherapy - "Наукові дослідження"
2. Click "Додати +" - a new expanded entry appears at the bottom, focused, with counters 0/150 and 0/1000; the button becomes disabled
3. Fill in both fields - "Додати +" become active again
4. In "Назва" type two spaces in a row - they collapse into one; a leading space is not accepted
5. Type 9 characters plus a space and click away - the value is trimmed and the minimum-length error appears

## CHECK LIST

- [x] New tests were added or existing were modified
- [ ] Changes have severe impact on users
- [ ] Changes have medium impact on users
- [x] Changes have light impact on users

## Summary by CodeRabbit

- **New Features**
  - Added the ability to create scientific references with generated identifiers.
  - New references open automatically for editing.
  - Add controls are disabled when existing references have incomplete or invalid details.
  - Reference names are normalized while typing and trimmed before validation.
  - Textarea inputs can now normalize content while preserving the typing position.

- **Bug Fixes**
  - Improved textarea behavior when external values change.
  - Ignored expand/collapse actions for references no longer in the list.
  - Preserved newlines while normalizing spaces in reference names.

### DIFF
--- a/src/components/admin/textarea-with-character-limit/TextAreaWithCharacterLimit.tsx
+++ b/src/components/admin/textarea-with-character-limit/TextAreaWithCharacterLimit.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef, useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react';
+import React, { forwardRef, useCallback, useEffect, useRef, useState } from 'react';
 import cn from 'classnames';
 import { ReactComponent as RemoveIcon } from '@/assets/icons/remove-query.svg';
 import { useInputWithCharacterLimit } from '@/hooks/admin/use-input-with-character-limit/useInputWithCharacterLimit';
@@ -21,6 +21,7 @@ export interface TextAreaWithCharacterLimitProps {
     onWarningChange?: (warning: string | null) => void;
     autoGrow?: boolean;
     maxRows?: number;
+    normalizeValue?: (value: string) => string;
 }
 
 export const TextAreaWithCharacterLimit = forwardRef<HTMLTextAreaElement, TextAreaWithCharacterLimitProps>(
@@ -42,16 +43,11 @@ export const TextAreaWithCharacterLimit = forwardRef<HTMLTextAreaElement, TextAr
             onWarningChange,
             autoGrow = false,
             maxRows = 10,
+            normalizeValue,
         },
         ref,
     ) => {
         const [localValue, setLocalValue] = useState(value ?? '');
-
-        const valueRef = useRef(value);
-
-        useLayoutEffect(() => {
-            valueRef.current = value;
-        }, [value]);
 
         useEffect(() => {
             setLocalValue(value ?? '');
@@ -85,11 +81,19 @@ export const TextAreaWithCharacterLimit = forwardRef<HTMLTextAreaElement, TextAr
             });
 
         const onInternalChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
-            setLocalValue(e.target.value);
+            const rawValue = e.target.value;
+            const nextValue = normalizeValue ? normalizeValue(rawValue) : rawValue;
+
+            if (normalizeValue && nextValue !== rawValue) {
+                const rawCaret = e.target.selectionStart ?? rawValue.length;
+                const nextCaret = normalizeValue(rawValue.slice(0, rawCaret)).length;
+
+                e.target.value = nextValue;
+                e.target.setSelectionRange(nextCaret, nextCaret);
+            }
+
+            setLocalValue(nextValue);
             handleChange(e);
-            Promise.resolve().then(() => {
-                setLocalValue(valueRef.current ?? '');
-            });
         };
 
         useEffect(() => {
--- a/src/components/admin/textarea-with-character-limit/TextAreaWithCharacterLimit.tsx
+++ b/src/components/admin/textarea-with-character-limit/TextAreaWithCharacterLimit.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef, useCallback, useEffect, useRef, useState } from 'react';
+import React, { forwardRef, useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react';
 import cn from 'classnames';
 import { ReactComponent as RemoveIcon } from '@/assets/icons/remove-query.svg';
 import { useInputWithCharacterLimit } from '@/hooks/admin/use-input-with-character-limit/useInputWithCharacterLimit';
@@ -47,6 +47,12 @@ export const TextAreaWithCharacterLimit = forwardRef<HTMLTextAreaElement, TextAr
     ) => {
         const [localValue, setLocalValue] = useState(value ?? '');
 
+        const valueRef = useRef(value);
+
+        useLayoutEffect(() => {
+            valueRef.current = value;
+        }, [value]);
+
         useEffect(() => {
             setLocalValue(value ?? '');
         }, [value]);
@@ -81,6 +87,9 @@ export const TextAreaWithCharacterLimit = forwardRef<HTMLTextAreaElement, TextAr
         const onInternalChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
             setLocalValue(e.target.value);
             handleChange(e);
+            Promise.resolve().then(() => {
+                setLocalValue(valueRef.current ?? '');
+            });
         };
 
         useEffect(() => {
--- a/src/pages/admin/hippotherapy-page/components/sections/scientific-references-section/ScientificReferencesSection.test.tsx
+++ b/src/pages/admin/hippotherapy-page/components/sections/scientific-references-section/ScientificReferencesSection.test.tsx
@@ -286,22 +286,6 @@ describe('ScientificReferencesSection', () => {
         expect(getAddButton()).toBeEnabled();
     });
 
-    it('collapses consecutive spaces in the name while typing', () => {
-        renderComponent();
-
-        fireEvent.change(screen.getByTestId('name-input-ref-1'), {
-            target: { value: '  Hello   world ' },
-        });
-
-        expect(mockOnChange).toHaveBeenCalledWith(
-            expect.objectContaining({
-                scientificReferences: expect.arrayContaining([
-                    expect.objectContaining({ localId: 'ref-1', name: 'Hello world ' }),
-                ]),
-            }),
-        );
-    });
-
     it('ignores a toggle for a reference that is no longer in the list', () => {
         renderComponent();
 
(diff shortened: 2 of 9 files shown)


Zest005:
The space-normalisation expression already existed inline in `TranslateHistorySectionForm.tsx` and `history-translation-schema.ts`. I moved it into `text-formatters.ts` instead of adding a third copy, but did not refactor those two call sites because it seems like a separate task to me

src/components/admin/textarea-with-character-limit/TextAreaWithCharacterLimit.tsx
```diff
@@ -81,6 +87,9 @@ export const TextAreaWithCharacterLimit = forwardRef<HTMLTextAreaElement, TextAr
         const onInternalChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
             setLocalValue(e.target.value);
             handleChange(e);
+            Promise.resolve().then(() => {
+                setLocalValue(valueRef.current ?? '');
+            });
```
fhyhtyh: This microtask forces localValue back to the prop after every keystroke. When an edit normalizes to a string equal to the current value (typing a second consecutive space, or a leading space), the [value] effect doesn't fire, so this runs and rewrites textarea.value. React doesn't restore the selection when focus hasn't moved, so the caret jumps to the end. Also: this assumes the parent updates value synchronously — a debounced/derived consumer would get the keystroke reverted to the stale prop. Can we only reconcile when the prop actually changed, rather than unconditionally after each change?

src/pages/admin/hippotherapy-page/components/sections/scientific-references-section/ScientificReferencesSection.tsx
```diff
@@ -28,6 +30,17 @@ const ScientificReferencesSectionComponent = ({ value, onChange, disabled }: Sci
...
     const [expandedIds, setExpandedIds] = useState<Set<string>>(new Set());
+    const [newReferenceLocalId, setNewReferenceLocalId] = useState<string | null>(null);
+
+    const areAllReferencesValid = useMemo(
+        () =>
+            value.scientificReferences.every(
+                (reference) =>
+                    !HIPPOTHERAPY_PAGE_VALIDATION_FUNCTIONS.validateText(reference.name) &&
+                    !HIPPOTHERAPY_PAGE_VALIDATION_FUNCTIONS.validateText(reference.url),
+            ),
```
fhyhtyh: validateText runs on the raw value here, but handleNameBlur / handleUrlBlur now validate trimed values. A name like "Reference " (trailing space → 10 chars) keeps the Add button enabled while the blur handler renders "Не менше 10 символів" for the trimmed 9-char string. isHippotherapyPageContentValid gates Publish with the same untrimmed check. Suggest trimming here (and in the publish gate) so all three agree.

src/pages/admin/hippotherapy-page/components/sections/scientific-references-section/ScientificReferencesSection.tsx
```diff
@@ -60,14 +90,16 @@ const ScientificReferencesSectionComponent = ({ value, onChange, disabled }: Sci
     };
 
     const handleNameChange = (localId: string, name: string) => {
+        const normalisedName = getNormalizedInputTextWhileTyping(name);
```
fhyhtyh: getNormalizedInputTextWhileTyping collapses \s+, which includes \n, and it's applied on every change here. The citation-name field renders as a multi-row auto-growing textarea, so this silently converts Enter into a space while the user types (previously raw text was kept until blur). If names are meant to be single-line, consider blocking Enter instead of turning it into a space; if multi-line is allowed, the normalization needs to preserve newlines.

Zest005: Fixed. The regex now only collapses literal spaces, so newlines are preserved.

src/pages/admin/hippotherapy-page/components/sections/scientific-references-section/ScientificReferencesSection.tsx
Zest005: Fixed. Extracted validateReferenceText, which trims before validating, and used it in areAllReferencesValid, both change handlers and both blur handlers.
HippotherapyPageSchema.ts (where it is isHippotherapyPageContentValid) is being rewritten right now in #581, so I'll check it after it merges.

src/components/admin/textarea-with-character-limit/TextAreaWithCharacterLimit.tsx
Zest005: Yeah, in InputWithCharacterLimit no parent ever modifies the value, so the re-sync never rewrites anything. I think I have two options:
1. Move the normalisation into the component behind an optional normalizeValue prop - it owns the DOM node, so it can restore the caret, and the microtask goes away entirely
2. Normalise on blur only and revert this change - simpler, but drops the wiki rule about collapsing spaces in real time and this differs from the [Wiki](https://github.com/ita-social-projects/VictoryCenter-Back/wiki/Space-management).

Which one is better?

src/pages/admin/hippotherapy-page/components/sections/scientific-references-section/ScientificReferencesSection.tsx
```diff
@@ -28,6 +30,17 @@ const ScientificReferencesSectionComponent = ({ value, onChange, disabled }: Sci
...
+    const [newReferenceLocalId, setNewReferenceLocalId] = useState<string | null>(null);
+
+    const areAllReferencesValid = useMemo(
+        () =>
+            value.scientificReferences.every(
+                (reference) =>
+                    !HIPPOTHERAPY_PAGE_VALIDATION_FUNCTIONS.validateText(reference.name) &&
+                    !HIPPOTHERAPY_PAGE_VALIDATION_FUNCTIONS.validateText(reference.url),
+            ),
+        [value.scientificReferences],
```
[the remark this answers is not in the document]
Zest005: Fixed. Extracted validateReferenceText, which trims before validating, and used it in areAllReferencesValid, both change handlers and both blur handlers.

src/components/admin/textarea-with-character-limit/TextAreaWithCharacterLimit.tsx
fhyhtyh: I think option 2 would be a better approach but you should probably clarify this with Olena first

Zest005: Olena replied: collapsing after input is fine, but she wants the extra spaces not to count towards the character limit.
That rules out option 2 on its own: with blur-only normalisation the counter and maxLength still work on the raw string while typing. Option 1 covers both - the value is already collapsed, so the counter and the limit are correct, and the component restores the caret itself.
So I'd go with option 1.

src/pages/admin/hippotherapy-page/components/sections/scientific-references-section/ScientificReferencesSection.tsx
```diff
@@ -78,23 +110,23 @@ const ScientificReferencesSectionComponent = ({ value, onChange, disabled }: Sci
...
+            setUrlErrors((prev) => ({ ...prev, [localId]: validateReferenceText(url) }));
         }
     };
 
     const handleNameBlur = (localId: string) => {
         const reference = value.scientificReferences.find((item) => item.localId === localId);
         setNameErrors((prev) => ({
             ...prev,
-            [localId]: HIPPOTHERAPY_PAGE_VALIDATION_FUNCTIONS.validateText(reference?.name ?? ''),
+            [localId]: validateReferenceText(reference?.name ?? ''),
```
mehalyna, line 118: The blur validation checks reference?.name untrimmed, allowing "ab " to pass the minimum-length check. Validate the trimmed name to ensure consistent behavior with space normalization rules.

Zest005: Already handled - validateReferenceText (line 27) trims before validating, so 'ab ' fails the check. Both blur handlers, both change handlers and areAllReferencesValid go through it.

src/components/admin/textarea-with-character-limit/TextAreaWithCharacterLimit.tsx
Zest005: Done

src/pages/admin/hippotherapy-page/components/sections/scientific-references-section/ScientificReferencesSection.test.tsx
```diff
@@ -247,4 +253,71 @@ describe('ScientificReferencesSection', () => {
...
+    it('keeps the add button disabled when a value only passes validation thanks to trailing spaces', () => {
+        renderComponent({
+            value: {
+                ...defaultValue,
+                scientificReferences: [{ localId: 'ref-1', id: 1, name: 'ab', url: 'https://example.com/one' }],
+            },
+        });
+
+        expect(getAddButton()).toBeDisabled();
+    });
```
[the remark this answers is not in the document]
Zest005: Done. 'ab ' is still invalid before trimming, so it wouldn't prove anything either. Used 'ab' + 9 spaces instead.

mehalyna (approved):

fhyhtyh (approved):

stkossman (approved):

